PR #78, opened 2015-01-18
added 404 to list of status codes inaccessible by head

shreyas-satish:


jace:
We might as well remove the HEAD request then. Checking for 404 is more or less the entire point of this validator.

shreyas-satish:
Sure, updated PR.

PR #83, opened 2015-01-25
Sidebar layout

Praseetha-KR:


jace:
Closing this for now since we've decided to use top navigation instead of sidebar navigation. However, sidebar may be required in future, so I'm keeping the branch for future use.

PR #86, opened 2015-03-04
Multiselect

vidya-ram:
Multiselect library for filter UI


shreyas-satish:
This was intended for filters in Hasjob, but we're not using multiselect anymore.

PR #89, opened 2015-04-06
Hgnav changes

vidya-ram:
Provision to add more links(eg: drafts) in dropdown menu of navigation bar.
Added background color while hovering over the items (hasgeek links in dropdown menu).


vidya-ram:
Move all the css style from components.html to corresponding stylesheet.

PR #90, opened 2015-04-27
Hgnav changes

vidya-ram:
Changes made to hgnav for adding extra links( like drafts ) to navbar. 	
Added hover effect for hasgeek site links( block-dropdown list items ). 	
Moved the css style from components.html to networkbar.scss.


baseframe/templates/baseframe/components.html
```diff
@@ -163,169 +163,7 @@
...
-    ul.hg-subnav {
-      margin-bottom: 0;
-    }
-
-    ul.hg-subnav > li > a {
-      padding: 12px 15px;
-    }
-  </style>
-  
+{% macro hgnav(site_title="HasGeek", site_nav=none, hasgeek_auth=true, hasgeek_network=true, links=[]) %}
```
jace: Let's not use `hasgeek_` here. `auth` and `network` should be sufficient.

baseframe/templates/baseframe/components.html
```diff
@@ -370,12 +208,15 @@
...
-{% macro add_hasgeek_auth() %}
+{% macro add_hasgeek_auth(links) %}
   {% if g.user -%}
     <a href="#" data-toggle="dropdown" class="dropdown-toggle" id="hg-user-btn">
       <i class="fa fa-user"></i><b class="caret"></b>
     </a>
     <ul class="dropdown-menu pull-right">
+      {%- for link in links %}
+        <li><a href="{{ link.url }}" {%- if link.rel %} rel="{{ link.rel }}"{% endif %} {%- if link.class %} class="{{ link.class }}"{% endif %}>{{ link.title }}</a></li>
+      {%- endfor %}
```
jace: This assumes links are only visible if logged in. Is there a use case for non-logged in links?

vidya-ram: This has been added for "Drafts" when logged in.
In case of non-logged there will be only log in button. 
Other Site related links are under site_nav

baseframe/static/css/networkbar.css
```diff
@@ -713,3 +713,117 @@
...
+#hgnav .nav {
+  margin-bottom: 0;
+}
+#hgnav .nav > li {
+  line-height: 36px;
+  color: #777;
+}
+#hgnav .nav > li > a {
+  padding: 0 15px;
+  color: #777;
```
shreyas-satish: The button [used](https://github.com/hasgeek/baseframe/blob/hgnav-changes/baseframe/templates/baseframe/components.html#L238) is `btn-primary`, so the text needs to be in a light color, `#fff` would do.

baseframe/templates/baseframe/components.html
```diff
@@ -163,204 +163,38 @@
...
-        </ul>
       </div>
-      <div class="navbar-header pull-right">
+      <div class="navbar-header pull-right hg-header-nav">
         <ul class="nav pull-right">
           <li class="dropdown pull-left">
-            {% if hasgeek_auth %}
-              {{add_hasgeek_auth()}}
+            {% if auth %}
+              {{add_hasgeek_auth(links)}}
```
jace: As per our coding convention, Jinja2 tags should have a single space padding inside, so this should be `{{ add_hasgeek_auth(links) }}`.

PR #92, opened 2015-05-22
Add provision to change login button text for lastuser and changed heading size.

vidya-ram:


vidya-ram:
@jace Could you please review this. Network bar changes

PR #108, opened 2015-11-24
Revert CodeMirror to 4.11.0

sindhus:


baseframe/Makefile
```diff
@@ -1,9 +1,9 @@
 # Makefile for tinymce
 all: tinymce codemirror
 
-tinymce3_zip = tinymce_3.5.11_jquery.zip
 tinymce4_zip = tinymce_4.3.0_jquery.zip
-codemirror_ver = 5.8.0
+tinymce4_zip = tinymce_4.1.7_jquery.zip
```
jace: This shouldn't be here.

sindhus: When I pulled and rebased from master with `git pull --rebase baseframe master`, `codemirror_ver` was right in between tinymce4 and tinymce5, so I moved the variable below which is why git shows it as a new addition.

jace: Look again. You've removed TinyMCE 3 and inserted a second copy for 4.

sindhus: Ok will fix.

PR #117, opened 2016-01-25
Added Fira 4

shreyas-satish:


jace:
Please break down the CSS into individual files. Some browsers download all fonts that are specified in the CSS, even if they are not used.

shreyas-satish:
Discussion moved to https://github.com/hasgeek/baseframe/pull/118

PR #119, opened 2016-01-28
Remove pyIsEmail's version; not required anymore

sindhus:


jace:
Since we are using a `>=` operator, this can stay. Our problem is only when using `==` or `<=`.

PR #131, opened 2017-01-26
rm unnecessary line

shreyas-satish:
Flask-WTF 0.14 has [deprecated](http://flask-wtf.readthedocs.io/en/stable/changelog.html) `csrf_enabled` and will remove this entirely in `1.0`.

### DIFF
--- a/baseframe/forms/fields.py
+++ b/baseframe/forms/fields.py
@@ -617,7 +617,6 @@ class FormField(wtforms.FormField):
     """
     def process(self, *args, **kwargs):
         retval = super(FormField, self).process(*args, **kwargs)
-        self.form.csrf_enabled = False
         del self.form.csrf_token
         return retval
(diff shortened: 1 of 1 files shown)


baseframe/forms/fields.py
```diff
@@ -617,7 +617,6 @@ class FormField(wtforms.FormField):
     """
     def process(self, *args, **kwargs):
         retval = super(FormField, self).process(*args, **kwargs)
-        self.form.csrf_enabled = False
         del self.form.csrf_token
```
jace: Is the `del` still necessary? Does FormField work if you don't initiate without CSRF disabled?

shreyas-satish: Without deleting the field, the label for `csrf_token` i.e `CSRF Token` shows up.

PR #132, opened 2017-02-13
Update select2 version to 4.0.3

### LINKED ISSUE #85: Upgrade Select2 to 4.0
Select2 has a 4.0 release in beta. This is a breaking release that no longer works with comma-separated values in hidden fields, so Baseframe's UserSelectField, GeonameField and others will need to be rewritten.

vidya-ram:
1. Add select2 4.0.3 version in which this issue https://github.com/select2/select2/issues/279
has been resolved.
2. Add check for data in pre_validate of DateTimeField.

Note:- Need to update hasgeek.tv

Fixes #85


vidya-ram:
Need to add support to provide old version of select2.

jace:
@vidya-ram It may be simpler to just update everywhere at once, instead of bothering with maintaining the old version.

jace:
Apps that use Select2: Lastuser, Hasjob, HasGeek TV.

jace:
I'll recommend removing select2 from the base dependency unless explicitly requested by the client app. Let's update all the client apps to specify which select2 they want.

jace:
We also need to update Baseframe's autocomplete widgets to work with the latest Select2, if they aren't already compatible.

jace:
This will fix #85 when closed, but only if the commit message mentions it.

baseframe/assets.py
```diff
@@ -87,8 +87,15 @@
...
-assets['select2.css'][Version('3.3.2')] = 'baseframe/css/select2.css'
+assets['select2.js'][Version('4.0.3')] = ('jquery.js>=1.7.2', 'baseframe/js/select2.js')
+assets['select2.css'][Version('4.0.3')] = 'baseframe/css/select2.css'
+assets['bootstrap-select2.css'][Version('0.1.0')] = 'baseframe/css/bootstrap-select2.css'
+assets['select2-baseframe.css'][Version('4.0.3')] = 'baseframe/css/select2-baseframe.css'
+assets['baseframe-select2.css'][Version('4.0.3')] = {'requires': [
+    'select2.css',
+    'bootstrap-select2.css',
+    'select2-baseframe.css'
+]}
```
jace: This is a bit confusing. Instead:

```python
assets['select2.css'][Version('4.0.3')] = 'baseframe/css/select2.css'
assets['select2-bootstrap.css'][Version('0.1.0')] = ('select2.css', 'baseframe/css/bootstrap-select2.css')
assets['select2-baseframe.css'][Version('4.0.3')] = ('select2-bootstrap.css', 'baseframe/css/select2-baseframe.css')
assets['select2-baseframe.js'][Version('4.0.3')] = {'requires': ['select2.js==4.0.3']}
```

baseframe/forms/fields.py
```diff
@@ -305,39 +305,12 @@ def process_formdata(self, valuelist):
         self._timezone_converted = False
 
     def pre_validate(self, form):
-        if self._timezone_converted is False:
+        if self.data and self._timezone_converted is False:
```
jace: This is an unrelated change. Why is this here?

baseframe/forms/widgets.py
```diff
@@ -36,6 +36,23 @@ def __call__(self, field, **kwargs):
...
+class Select2Widget(Select):
+    """
+    Add support of choices with ``optgroup`` to the ``Select`` widget.
+    """
+    def __call__(self, field, **kwargs):
+        kwargs.setdefault('id', field.id)
+        kwargs.pop('type', field.type)
+        if field.multiple:
+            kwargs['multiple'] = 'multiple'
+        html = ['<select %s>' % html_params(name=field.name, **kwargs)]
```
jace: This should have a `class="select2"` so that we don't clobber other select inputs. Add it via kwargs just in case there is an existing class argument.

jace: WTForms convention is to use the term `class_` for CSS class. See our other widgets for how we handle the ambiguity.

baseframe/assets.py
vidya-ram: Fixed

baseframe/forms/widgets.py
vidya-ram: Fixed

baseframe/forms/fields.py
vidya-ram: Incase DateTimeField is optional in the form, it was failing. So, I added the check. Do I remove it now and add it as different PR?

PR #152, opened 2017-05-03
autoreload templates

shreyas-satish:


jace:
If Jinja2 turned off auto reload for performance reasons, turning it on again may not be a good idea. The config flag should be in `development.py`.

PR #155, opened 2017-07-17
Fix select2 for single select option.

vidya-ram:


baseframe/forms/fields.py
```diff
@@ -424,6 +438,20 @@ class AutocompleteField(AutocompleteFieldBase, StringField):
     multiple = False
     widget = Select2Widget()
 
+    def _value(self):
+        if self.data:
+            return self.data
```
jace: Shouldn't this be `self.data.userid` as well?

PR #158, opened 2017-08-04
Added tests for ValidURL and fixed a minor bug

shreyas-satish:


baseframe/forms/validators.py
```diff
@@ -238,8 +238,9 @@ def check_url(self, invalid_urls, url, text=None):
                     # For text patterns, do a substring search. For regex patterns (assumed so if not text),
                     # do a regex search. Test with the final URL from the response, after redirects,
                     # but report errors using the URL the user provided
-                    if (isinstance(pattern, basestring) and pattern in rurl) or (pattern.search(rurl) is not None):
-                        return message.format(url=url, text=text)
+                    if (isinstance(pattern, basestring) and pattern in rurl) or (
+                        isinstance(pattern, re._pattern_type) and pattern.search(rurl) is not None):
```
jace: What broke to introduce this change? Referring to underscore-prefixed symbols in modules maintained by someone else is not recommended.

shreyas-satish: In case the pattern is a string, and not in `rurl`, the conditional fails since it attempts to call `search` on a string.

shreyas-satish: Works if the type is checked instead? Ex: `isinstance(pattern, type(re.compile(r''))`)

jace: Try `if pattern in rurl if isinstance(pattern, basestring) else pattern.search(rurl) is not None:`

shreyas-satish: Removed the type check, broke the conditional ^ into an `if/elif` for clarity.

baseframe/forms/validators.py
```diff
@@ -238,7 +238,10 @@ def check_url(self, invalid_urls, url, text=None):
                     # For text patterns, do a substring search. For regex patterns (assumed so if not text),
                     # do a regex search. Test with the final URL from the response, after redirects,
                     # but report errors using the URL the user provided
-                    if (isinstance(pattern, basestring) and pattern in rurl) or (pattern.search(rurl) is not None):
+                    if isinstance(pattern, basestring):
+                        if pattern in rurl:
+                            return message.format(url=url, text=text)
+                    elif pattern.search(rurl) is not None:
                         return message.format(url=url, text=text)
```
jace: Same `return` statement twice? The other syntax didn't work?

shreyas-satish: The other syntax works, but this seems a little clearer. Would you prefer the other syntax?

PR #161, opened 2017-08-08
Added tests for SQLAlchemy validators, made model mandatory in AvailableAttr

shreyas-satish:


tests/fixtures.py
```diff
@@ -5,6 +5,13 @@
 from baseframe import _, __
 import baseframe.forms as forms
 
+app1 = Flask(__name__)
+app1.config['SQLALCHEMY_DATABASE_URI'] = 'sqlite://'
+app1.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = False
+app2 = Flask(__name__)
+app2.config['SQLALCHEMY_DATABASE_URI'] = 'postgresql:///baseframe_test'
```
jace: Three slashes?

tests/test_sqlalchemy.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class ContainerForm(forms.Form):
+    name = forms.StringField(__("Name"), validators=[forms_sqlachemy.AvailableName()])
+    title = forms.StringField(__("Title"), validators=[forms_sqlachemy.AvailableAttr('title')])
+    content = forms.TextAreaField(__("Content"))
+
+
+class FormSQLAlchemyTestCase(BaseframeTestCase):
+    def setUp(self):
```
jace: Add `app = app1` here and use `self.app` in `setUp`. You can make a subclass below that substitutes with `app = app2` to re-run tests on the other database.

jace: Don't inherit from `BaseframeTestCase` as that messes with `self.app`. Just add the `init_app` call here.

tests/fixtures.py
shreyas-satish: My development config for PostgreSQL is specified slightly differently (with the username and port). So I borrowed the config from [coaster](https://github.com/hasgeek/coaster/blob/master/tests/test_models.py#L26).

shreyas-satish: Three slashes seems to be the correct syntax, two slashes errors out.

baseframe/forms/sqlalchemy.py
```diff
@@ -25,6 +25,8 @@ def __init__(self, attr, message=None, model=None):
 
     def __call__(self, form, field):
         model = self.model or form.edit_model
+        if not model:
+            raise TypeError(u"Either the field or the form MUST be linked to a model")
```
jace: `self` is the validator, not the field. Either the validator or the form must be linked to a model.

PR #174, opened 2017-11-24
added config loader for webpack and view helper for hashed assets

shreyas-satish:


baseframe/views.py
```diff
@@ -85,11 +85,18 @@ def ext_assets(assets):
         return gen_assets_url(assets)
 
 
+def asset_path(bundle_key):
+    return '{path}/{bundle}'.format(
+    path=current_app.config['ASSET_BASE_PATH'],
+    bundle=current_app.config['ASSETS'][bundle_key])
```
jace: This will raise a `KeyError` that will propagate to the template. Is that acceptable? (It may or may not be, depending on expected usage.)

PR #176, opened 2017-12-14
Add support to send html form for ajax requests

vidya-ram:
1. For HGTV SPA to reuse `render_form` so that the forms doesn't have to be rewritten on the client side, add support to `render_form` to send html form. Adding autoform_xhr.html.jinja2 for this.
2. Add an empty div field to each of the form fields so that errors received from the backend can be rendered in these on the client side.
3. Fix tinymce, 
a) to cleanly remove an editor instance and avoid errors
https://stackoverflow.com/questions/4651676/how-do-i-remove-tinymce-and-then-re-add-it
b) to sync changes to textfield
https://stackoverflow.com/questions/2122085/jquery-and-tinymce-textarea-value-doesnt-submit


baseframe/forms/auto.py
```diff
@@ -17,25 +18,28 @@ class ConfirmDeleteForm(Form):
...
 
-def render_form(form, title, message='', formid='form', submit=__(u"Submit"), cancel_url=None, ajax=False):
+def render_form(form, title, message='', formid='form', submit=__(u"Submit"), cancel_url=None, ajax=False, with_chrome=True, error_template=False):
     multipart = False
     for field in form:
         if isinstance(field.widget, wtforms.widgets.FileInput):
             multipart = True
     if form.errors:
+        # TODO: Why is this 200?
         code = 200  # 400
```
jace: Because HTTP 400 breaks `$.ajax`. Instead of rendering the new response with errors, it was firing the client-side error handler, which (depending on the app's code) says the server couldn't be contacted.

baseframe/forms/auto.py
```diff
@@ -54,18 +58,26 @@ def render_redirect(url, code=302):
         return redirect(url, code=code)
 
 
-def render_delete_sqla(obj, db, title, message, success=u'', next=None, cancel_url=None):
+def render_delete_sqla(obj, db, title, message, success=u'', next=None, cancel_url=None, with_response=True):
     if not obj:
         abort(404)
     form = ConfirmDeleteForm()
     if request.method in ('POST', 'DELETE') and form.validate():
+        print ('form', request.form)
```
jace: To be removed.

baseframe/forms/auto.py
```diff
@@ -54,18 +58,26 @@ def render_redirect(url, code=302):
...
         if 'delete' in request.form or request.method == 'DELETE':
             db.session.delete(obj)
             db.session.commit()
             if success:
                 flash(success, 'success')
+            if request.is_xhr:
+                return make_response(jsonify(status='ok', doc=_(u"Deleted"), result={'url': next or url_for('index')}), 303)
```
jace: 303 response needs a `Location` header. The Ajax handler will replay a GET request to the `Location` URL. Better to respond with 200 here. [This discussion argues for 200](https://stackoverflow.com/questions/6439416/deleting-a-resource-using-http-delete).

baseframe/forms/auto.py
```diff
@@ -54,18 +58,26 @@ def render_redirect(url, code=302):
...
             return render_redirect(next or url_for('index'), code=303)
         else:
+            if request.is_xhr:
+                return make_response(jsonify(status='ok', doc=_(u"Cancel"), result={'url': cancel_url or next or url_for('index')}), 303)
```
jace: 200 here too.

baseframe/forms/auto.py
```diff
@@ -54,18 +58,26 @@ def render_redirect(url, code=302):
...
             return render_redirect(cancel_url or next or url_for('index'), code=303)
-    template = THEME_FILES[current_app.config['theme']]['delete.html.jinja2']
-    return make_response(render_template(template, form=form, title=title, message=message))
+    if with_response:
+        template = THEME_FILES[current_app.config['theme']]['delete.html.jinja2']
+        return make_response(render_template(template, form=form, title=title, message=message))
+    template = THEME_FILES[current_app.config['theme']]['delete_template.html.jinja2']
+    return render_template(template, form=form, title=title, message=message)
```
jace: I don't get the difference between `delete.html.jinja2` and `delete_template.html.jinja2`. Could they have more descriptive names?

baseframe/templates/baseframe/mui/forms.html.jinja2
```diff
@@ -38,8 +37,9 @@
         <div>
           <div class="mui-checkbox">
             <label>
-              {{ rendererrors(field) }}
               {{ field | render_field_options(class="field-" + field.id + " " + widget_css_class, tabindex=tabindex, rows=rows) }} {{ field.label.text }}
+              {{ rendererrors(field) }}
+              {%- if error_template %}<p class="mui-form--error" id="{{field.name}}-error"></p>{%- endif %}
```
jace: Shouldn't the `rendererrors` macro do this? Is `error_template` a parameter to be passed only for the MUI theme?

baseframe/templates/baseframe/mui/forms.html.jinja2
```diff
@@ -193,6 +197,9 @@
   <script type="text/javascript">
     $(function() {
   {%- endif %}
+      if (typeof(tinyMCE) != 'undefined') {
+        tinymce.remove('textarea#{{ field.id }}');
+      }
```
jace: Why was this change needed?

vidya-ram: Tinymce was failing to load the second time since there isn't a page reload in case of hgtv spa.
So, this line takes care of cleanly removing any previous instance tinymce.

baseframe/forms/auto.py
vidya-ram: Fixed

baseframe/forms/auto.py
vidya-ram: Fixed

baseframe/forms/auto.py
vidya-ram: Fixed

baseframe/templates/baseframe/mui/forms.html.jinja2
vidya-ram: Since the form is rendered client side on hgtv spa, this empty p field was added so that client can render the field errors just below it.

baseframe/forms/auto.py
vidya-ram: Removing it for now. Could we look at `render_message` and `render_delete_sqla` as part crud-view PR?

baseframe/templates/baseframe/mui/forms.html.jinja2
jace: I'm not comfortable with this. This is a render flag that has to be set in business logic. We can't be passing around an `error_template` parameter. The code that displays errors has to use some other mechanism to find the right place to display them.

baseframe/templates/baseframe/mui/forms.html.jinja2
```diff
@@ -287,6 +299,10 @@
   <script type="text/javascript">
     $(function() {
   {%- endif %}
+  // Activate select2.js for non-mobile browsers
+  if (!navigator.userAgent.match(/(iPod|iPad|iPhone|Android)/)) {
```
jace: We need a better test than this for mobile. Or at least for when select2 is activated.

shreyas-satish: This is an option - https://github.com/hgoebl/mobile-detect.js/

baseframe/static/js/baseframe-material.js
```diff
@@ -249,6 +249,22 @@ window.Baseframe.Forms = {
...
     })
+  },
+  showValidationErrors: function(errors) {
+    Object.keys(errors).forEach(function(fieldName) {
+      if (Object.prototype.toString.call(errors[fieldName]) === '[object Array]') {
+        // Create p and add error as its content
+        var errorElem = document.createElement('p');
+        errorElem.classList.add('mui-form--error');
+        errorElem.innerText = errors[fieldName];
+        var field = document.getElementById(fieldName);
```
shreyas-satish: This needs to be guaranteed to be unique.

baseframe/static/js/baseframe-material.js
```diff
@@ -249,6 +249,22 @@ window.Baseframe.Forms = {
         }
       }
     })
+  },
+  showValidationErrors: function(errors) {
+    Object.keys(errors).forEach(function(fieldName) {
+      if (Object.prototype.toString.call(errors[fieldName]) === '[object Array]') {
```
shreyas-satish: Specify the polyfill for `Array.isArray` in `baseframe.js`, so it can reused.

baseframe/templates/baseframe/mui/forms.html.jinja2
vidya-ram: Would detecting if the device supports touch events with `Modernizr.touch` suffice?

baseframe/static/js/baseframe-material.js
vidya-ram: Only IE8 doesn't support `Array.isArray` so I guess we could use it.

baseframe/static/js/baseframe-material.js
vidya-ram: Fixed

baseframe/forms/auto.py
```diff
@@ -17,25 +18,25 @@ class ConfirmDeleteForm(Form):
...
+    if not formid:
+        formid = 'form-' + buid()
     for field in form:
         if isinstance(field.widget, wtforms.widgets.FileInput):
             multipart = True
-    if form.errors:
-        code = 200  # 400
-    else:
-        code = 200
+    if not with_chrome:
```
shreyas-satish: What if the client wants a chrome-less ajax form? This condition implies that they would be returned `autoform` instead of `ajaxform`?

baseframe/static/js/baseframe-material.js
```diff
@@ -249,6 +249,26 @@ window.Baseframe.Forms = {
         }
       }
     })
+  },
+  showValidationErrors: function(formId, errors) {
+    Object.keys(errors).forEach(function(fieldName) {
+      if (Array.isArray(errors[fieldName])) {
+        var form = document.getElementById(formId);
+        var fieldWrapper = form.querySelector("#field-" + fieldName);
```
shreyas-satish: Why `querySelector` and not `getElementById`?

baseframe/static/js/baseframe-material.js
```diff
@@ -249,6 +249,26 @@ window.Baseframe.Forms = {
         }
       }
     })
+  },
+  showValidationErrors: function(formId, errors) {
+    Object.keys(errors).forEach(function(fieldName) {
+      if (Array.isArray(errors[fieldName])) {
```
shreyas-satish: Is there a chance that `errors[fieldName]` would _not_ be an array?

baseframe/static/js/baseframe-material.js
```diff
@@ -249,6 +249,26 @@ window.Baseframe.Forms = {
         }
       }
     })
+  },
+  showValidationErrors: function(formId, errors) {
```
shreyas-satish: Can you add a top-level comment that provides an overview of what this function does?

baseframe/forms/auto.py
vidya-ram: Since this template is just the form without the entire page just like the `ajaxform.html.jinja`. Changed it to use the same.
Fixed

baseframe/static/js/baseframe-material.js
vidya-ram: getElementById is not available here

baseframe/static/js/baseframe-material.js
vidya-ram: Yes, if the error is not due a form validation error.
In case of hgtv, it could be other errors in the form like logo upload error.

baseframe/static/js/baseframe-material.js
vidya-ram: Fixed

baseframe/forms/auto.py
```diff
@@ -17,25 +18,25 @@ class ConfirmDeleteForm(Form):
...
-        code = 200
-    if request.is_xhr and ajax:
+    if not with_chrome:
         template = THEME_FILES[current_app.config['theme']]['ajaxform.html.jinja2']
-        return make_response(render_template(template, form=form, title=title,
+        return render_template(template, form=form, title=title,
             message=message, formid=formid, submit=submit,
-            cancel_url=cancel_url, multipart=multipart), code)
+            cancel_url=cancel_url, ajax=ajax, multipart=multipart)
+    if request.is_xhr and ajax:
```
shreyas-satish: Why do we need to check for both `request.is_xhr` and `ajax`? `request.is_xhr` should suffice?

vidya-ram: This was check that was previously there.
@jace will just `request.is_xhr` suffice?

jace: There are two considerations here:

1. This is an Ajax request (`request.is_xhr` is True)
2. The developer wanted this to be an Ajax form (`ajax` is True)

The developer may not want an Ajax form for reasons including (a) the page being simple enough to afford a full reload, and (b) testing the business logic with the interactive debugger.

baseframe/static/js/baseframe-material.js
```diff
@@ -249,6 +249,43 @@ window.Baseframe.Forms = {
...
+    For each error, a 'p' tag is created if not present and
+    assigned the error value as its text content.
+    The field wrapper and field are queried in the DOM
+    using the unique form id. And the newly created 'p' tag
+    is inserted in the DOM below the field.
+  */
+  showValidationErrors: function(formId, errors) {
+    Object.keys(errors).forEach(function(fieldName) {
+      if (Array.isArray(errors[fieldName])) {
+        var form = document.getElementById(formId);
```
shreyas-satish: This need not run inside a loop, right? That way it can be fetched once before `Object.keys` is iterated on? Since it's going to be the same for all the errors.

baseframe/static/js/baseframe-material.js
```diff
@@ -249,6 +249,43 @@ window.Baseframe.Forms = {
...
+        var fieldWrapper = form.querySelector("#field-" + fieldName);
+        var errorElem = fieldWrapper.querySelector('.mui-form--error');
+        // If error P tag doesn't exist, create it
+        if(!errorElem) {
+          errorElem = document.createElement('p');
+          errorElem.classList.add('mui-form--error');
+        }
+        errorElem.innerText = errors[fieldName][0];
+        var field = form.querySelector("#" + fieldName)
```
shreyas-satish: Isn't there a chance of a collision here? Should this be prepended with `formId`?

baseframe/static/js/baseframe-material.js
vidya-ram: Fixed

baseframe/static/js/baseframe-material.js
vidya-ram: The field elements inside `form` are queried so there isn't a chance of collision

baseframe/static/js/baseframe-material.js
```diff
@@ -249,6 +249,45 @@ window.Baseframe.Forms = {
         }
       }
     })
+  },
+  /* Takes 'formId' and 'errors'
+     'formId' is the id attribute of the form for which errors needs to be displayed
+     'errors' is the wtf form validation errors expected in the following format
```
shreyas-satish: Typo: it's `WTForm` or `wtform`, not `wtf form`.

baseframe/static/js/baseframe-material.js
```diff
@@ -249,6 +249,45 @@ window.Baseframe.Forms = {
...
+    The field wrapper and field are queried in the DOM
+    using the unique form id. And the newly created 'p' tag
+    is inserted in the DOM below the field.
+  */
+  showValidationErrors: function(formId, errors) {
+    var form = document.getElementById(formId);
+    Object.keys(errors).forEach(function(fieldName) {
+      if (Array.isArray(errors[fieldName])) {
+        var fieldWrapper = form.querySelector("#field-" + fieldName);
+        if(fieldWrapper) {
```
shreyas-satish: Space after `if`.

baseframe/static/js/baseframe-material.js
```diff
@@ -249,6 +249,45 @@ window.Baseframe.Forms = {
...
+          var errorElem = fieldWrapper.querySelector('.mui-form--error');
+          // If error P tag doesn't exist, create it
+          if(!errorElem) {
```
shreyas-satish: Same.

baseframe/static/js/baseframe-material.js
vidya-ram: Fixed

baseframe/static/js/baseframe-material.js
vidya-ram: Fixed

baseframe/static/js/baseframe-material.js
vidya-ram: Fixed

baseframe/forms/auto.py
```diff
@@ -17,25 +18,24 @@ class ConfirmDeleteForm(Form):
...
             multipart = True
-    if form.errors:
-        code = 200  # 400
-    else:
-        code = 200
+    if not with_chrome:
+        template = THEME_FILES[current_app.config['theme']]['ajaxform.html.jinja2']
+        return render_template(template, form=form, title=title,
+            message=message, formid=formid, ref_id=ref_id, submit=submit,
+            cancel_url=cancel_url, ajax=ajax, multipart=multipart)
```
jace: What's the difference between this render and the one below?

baseframe/templates/baseframe/bootstrap3/autoform.html.jinja2
```diff
@@ -14,10 +14,10 @@
     $(function() {
       {{ widgetscripts(form, script=false) }}
       // Reset disabled state if page was reloaded (TODO: does this really get called?)
-      $('#{{ formid }} .submit-disabled').prop('disabled', false);
+      $('#{{ ref_id }} .submit-disabled').prop('disabled', false);
```
jace: `ref_id` will change here if this is after the reload. Test if this line ever gets called at all.

baseframe/static/js/baseframe-material.js
```diff
@@ -2,7 +2,7 @@
 
 function activate_widgets() {
   // Activate select2.js for non-mobile browsers
-  if (!navigator.userAgent.match(/(iPod|iPad|iPhone|Android)/)) {
+  if (!Modernizr.touch) {
```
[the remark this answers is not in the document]
vidya-ram: I tested it on mobile and native select dropdown seems to be more user friendly than select2

baseframe/templates/baseframe/mui/forms.html.jinja2
```diff
@@ -139,19 +139,19 @@
...
-<form data-parsley-validate id="{{ formid }}" method="POST" {%- if action %} action="{{ action }}" {%- endif %}{%- if multipart %} enctype="multipart/form-data" {%- endif %} accept-charset="UTF-8" class="mui-form mui-form--margins">
+{% macro renderform(form, formid, submit, ref_id='form', message='', action='', cancel_url='', multipart=false, style='horiz') %}
+<form data-parsley-validate id="{{ ref_id }}" method="POST" {%- if action %} action="{{ action }}" {%- endif %}{%- if multipart %} enctype="multipart/form-data" {%- endif %} accept-charset="UTF-8" class="mui-form mui-form--margins hg-form">
   {{ renderform_inner(form, formid or none, style=style) }}
   {{ rendersubmit([(none, submit or _("Submit"), 'mui-btn--primary')], cancel_url=cancel_url, style=style, csrf_error=form.csrf_token.errors) }}
 </form>
 {% endmacro %}
 
-{%- macro ajaxform(formid, request, force=false) %}
+{%- macro ajaxform(request, ref_id='form', force=false) %}
```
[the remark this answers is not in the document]
vidya-ram: Fixed

baseframe/static/js/baseframe.js
```diff
@@ -2,7 +2,7 @@
 
 function activate_widgets() {
   // Activate select2.js for non-mobile browsers
-  if (!navigator.userAgent.match(/(iPod|iPad|iPhone|Android)/)) {
+  if (!Modernizr.touch) {
```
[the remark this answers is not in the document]
vidya-ram: Fixed

baseframe/forms/auto.py
vidya-ram: Here we are passing just the form html to the client app to render.

baseframe/static/js/baseframe.js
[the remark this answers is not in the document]
vidya-ram: Fixed

baseframe/templates/baseframe/bootstrap3/autoform.html.jinja2
vidya-ram: Removed this line since it isn't being called

baseframe/templates/baseframe/mui/autoform.html.jinja2
```diff
@@ -14,9 +14,9 @@
     $(function() {
       {{ widgetscripts(form, script=false) }}
       // Reset disabled state if page was reloaded (TODO: does this really get called?)
-      $('#{{ formid }} .submit-disabled').prop('disabled', false);
+      $('#{{ ref_id }} .submit-disabled').prop('disabled', false);
```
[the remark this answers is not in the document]
vidya-ram: Removed this line since it isn't being called

PR #178, opened 2018-01-30
Baseframe form id

vidya-ram:
1. If client has passed a `formid`, prefix it with 'form-' and use it as id attribute for the form.
2. Incase the client doesn't pass a `formid`, generate one.

### DIFF
--- a/baseframe/forms/auto.py
+++ b/baseframe/forms/auto.py
@@ -18,9 +18,12 @@ class ConfirmDeleteForm(Form):
     cancel = SubmitField(__(u"Cancel"))
 
 
-def render_form(form, title, message='', formid='form', submit=__(u"Submit"), cancel_url=None, ajax=False, with_chrome=True):
+def render_form(form, title, message='', formid='', submit=__(u"Submit"), cancel_url=None, ajax=False, with_chrome=True):
     multipart = False
-    ref_id = 'form-' + buid()
+    if not formid:
+        ref_id = 'form-' + buid()
+    else:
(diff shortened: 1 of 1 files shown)


baseframe/forms/auto.py
```diff
@@ -18,9 +18,12 @@ class ConfirmDeleteForm(Form):
     cancel = SubmitField(__(u"Cancel"))
 
 
-def render_form(form, title, message='', formid='form', submit=__(u"Submit"), cancel_url=None, ajax=False, with_chrome=True):
+def render_form(form, title, message='', formid='', submit=__(u"Submit"), cancel_url=None, ajax=False, with_chrome=True):
```
jace: Use `None` instead of `''` for the default value.

baseframe/forms/auto.py
```diff
@@ -18,9 +18,12 @@ class ConfirmDeleteForm(Form):
...
     multipart = False
-    ref_id = 'form-' + buid()
+    if not formid:
+        ref_id = 'form-' + buid()
+    else:
+        ref_id = 'form-' + formid
```
jace: This could be `ref_id = 'form-' + (formid or buid())`

baseframe/forms/auto.py
vidya-ram: Fixed

baseframe/forms/auto.py
vidya-ram: Fixed

PR #182, opened 2018-02-14
Add ajax post handler to Baseframe

### LINKED ISSUE #179: Deprecate ajaxform
[ajaxform](https://github.com/hasgeek/baseframe/blob/master/baseframe/templates/baseframe/mui/forms.html.jinja2#L149) is currently a Jinja macro that wraps a JS function call to [`ajaxSubmit`](http://malsup.com/jquery/form/#api). This makes it unnecessarily complex for a client app's JS to register custom handlers that run before, on or after a form has been submitted.

I propose `ajaxform` in its current form be deprecated. `baseframe` should expose a new JS interface that processes AJAX requests given a URL and also accepts custom callbacks from a client app's JS.

Since ajaxform uses `ajaxSubmit`, it's a good opportunity to replace this with a more well maintained and a lighter library that does not depend jQuery. [axios](https://github.com/axios/axios) is an option.

vidya-ram:
Add `handleAjaxPost` function to `Baseframe.Forms`
This takes care of AJAX post requests for a client app given a URL, custom call backs for success & error scenarios.

Fixes #179

### DIFF
--- a/baseframe/static/js/baseframe-bootstrap.js
+++ b/baseframe/static/js/baseframe-bootstrap.js
@@ -298,14 +298,13 @@ window.Baseframe.Forms = {
    'formSelector' - Form selector to query the DOM for the form
    'onSuccess' - A callback function that is executed if the request succeeds
    'onError' - A callback function that is executed if the request fails
-   'config' -  An object that can contain data, contentType, dataType, 
-      beforeSend function
-    submitForm handles form submit, serializes the form values,
+   'config' -  An object that can contain data, contentType, dataType, beforeSend function
+    handleFormSubmit handles form submit, serializes the form values,
       disables the submit button to prevent double submit,
       displays the loading indicator and submits the form via ajax.
       On completing the ajax request, calls the onSuccess/onError callback function.
   */
-  submitForm: function(url, formSelector, onSuccess, onError, config) {
+  handleFormSubmit: function(url, formSelector, onSuccess, onError, config) {
     $(formSelector).find('button[type="submit"]').click(function(event) {
       event.preventDefault();
       $.ajax({
--- a/baseframe/static/js/baseframe-bootstrap.js
+++ b/baseframe/static/js/baseframe-bootstrap.js
@@ -254,8 +254,8 @@ window.Baseframe.Forms = {
       }
     })
   },
-  /* Takes 'formSelector' and 'errors'
-   'formSelector' is the form for which errors needs to be displayed
+  /* Takes 'formId' and 'errors'
+   'formId' is the id attribute of the form for which errors needs to be displayed
    'errors' is the WTForm validation errors expected in the following format
     {
       "title": [
@@ -271,8 +271,8 @@ window.Baseframe.Forms = {
     using the unique form id. And the newly created 'p' tag
     is inserted in the DOM below the field.
   */
-  showValidationErrors: function(formSelector, errors) {
-    var form = document.querySelector(formSelector);
+  showValidationErrors: function(formId, errors) {
+    var form = document.getElementById(formId);
     Object.keys(errors).forEach(function(fieldName) {
       if (Array.isArray(errors[fieldName])) {
         var fieldWrapper = form.querySelector("#field-" + fieldName);
--- a/baseframe/static/js/baseframe-bootstrap.js
+++ b/baseframe/static/js/baseframe-bootstrap.js
@@ -293,9 +293,9 @@ window.Baseframe.Forms = {
       }
     });
   },
-  /* Takes formSelector, url, onSuccess, onError, config
+  /* Takes formId, url, onSuccess, onError, config
+   'formId' - Form id selector to query the DOM for the form
    'url' - The url to which the post request is sent
-   'formSelector' - Form selector to query the DOM for the form
    'onSuccess' - A callback function that is executed if the request succeeds
    'onError' - A callback function that is executed if the request fails
    'config' -  An object that can contain dataType, beforeSend function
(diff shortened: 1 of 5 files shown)


baseframe/static/js/baseframe-bootstrap.js
```diff
@@ -292,6 +292,28 @@ window.Baseframe.Forms = {
         }
       }
     });
+  },
+  handleAjaxPost: function(config) {
```
shreyas-satish: This isn't a generic AJAX POST handler, right? Seems like it's a form handler. So, `submitForm` perhaps?

shreyas-satish: Can you please add a comment that explains what this function does and what attributes are excepted/accepted in `config`?

baseframe/static/js/baseframe-bootstrap.js
```diff
@@ -292,6 +292,28 @@ window.Baseframe.Forms = {
...
+      $.ajax({
+        url: config.url,
+        type: 'POST',
+        data: config.data ? config.data : $.param($(config.formSelector).serializeArray()),
+        contentType : config.contentType ? config.contentType : 'application/x-www-form-urlencoded; charset=UTF-8',
+        dataType: config.dataType ? config.dataType : 'json',
+        beforeSend: function() {
+          if (config.formSelector) {
+            $(config.formSelector).find('button[type="submit"]').prop('disabled', true);
+            $(config.formSelector).find(".loading").removeClass('hidden');
```
shreyas-satish: Good to allow for a `beforeSend` callback?

baseframe/static/js/baseframe-bootstrap.js
```diff
@@ -292,6 +292,28 @@ window.Baseframe.Forms = {
...
+          }
+        }
+      }).done(function (remoteData) {
+        config.onSuccess(config, remoteData);
```
shreyas-satish: Why is `config` passed back to the callback function? Trying to understand the use-case.

shreyas-satish: If `onSuccess` is a necessary argument, then it's perhaps better off as a function argument instead of a config variable.

baseframe/static/js/baseframe-bootstrap.js
```diff
@@ -292,6 +292,28 @@ window.Baseframe.Forms = {
         }
       }
     });
+  },
+  handleAjaxPost: function(config) {
+    $(config.formSelector).find('button[type="submit"]').click(function(event) {
```
shreyas-satish: Expecting an id such as `#form` ties it to jQuery-specific syntax. `getElementById` does not need a `#`, so removing the jQuery dependency in future will become trouble-some if it requires a `#` to be prepended to the selector.

baseframe/static/js/baseframe-bootstrap.js
vidya-ram: Yes, it is a generic ajax post handler. Not sure why it seemed like a form handler. 
'application/x-www-form-urlencoded; charset=UTF-8' is the default contentType in jquery ajax

baseframe/static/js/baseframe-bootstrap.js
vidya-ram: Since it is a callback and client wouldn't have the context.
https://github.com/hasgeek/boxoffice/blob/editform/boxoffice/assets/js/views/org.js#L76

baseframe/static/js/baseframe-bootstrap.js
shreyas-satish: > Not sure why it seemed like a form handler.

Because line 297 explicitly looks for a HTML element (likely a `form` element) with a `submit` button and adds a `click` handler to the button.

baseframe/static/js/baseframe-bootstrap.js
vidya-ram: It doesn't have to be `#form` it can be `.form` so even in the future if we replace this `document.querySelector` it should work

baseframe/static/js/baseframe-bootstrap.js
shreyas-satish: By not generic I meant to say that, I can't issue a plain JSON POST request without having a form element in my HTML.

vidya-ram: Fixed

vidya-ram: Added

baseframe/static/js/baseframe-bootstrap.js
vidya-ram: Fixed

baseframe/static/js/baseframe-bootstrap.js
vidya-ram: Changed

vidya-ram: Removed

baseframe/static/js/baseframe-bootstrap.js
```diff
@@ -292,6 +292,39 @@ window.Baseframe.Forms = {
         }
       }
     });
+  },
+  /* Takes formSelector, url, onSuccess, onError, config
+   'url' - The url to which the post request is sent
+   'formSelector' - Form selector to query the DOM for the form
+   'onSuccess' - A callback function that is executed if the request succeeds
+   'onError' - A callback function that is executed if the request fails
+   'config' -  An object that can contain data, contentType, dataType, beforeSend function
```
shreyas-satish: It can't contain `data` anymore. `beforeSend` callback.

baseframe/static/js/baseframe-bootstrap.js
```diff
@@ -271,8 +271,8 @@ window.Baseframe.Forms = {
     using the unique form id. And the newly created 'p' tag
     is inserted in the DOM below the field.
   */
-  showValidationErrors: function(formId, errors) {
-    var form = document.getElementById(formId);
+  showValidationErrors: function(formSelector, errors) {
+    var form = document.querySelector(formSelector);
```
shreyas-satish: This is a breaking change, no?

baseframe/templates/baseframe/bootstrap3/ajaxform.html.jinja2
```diff
@@ -1,6 +1,8 @@
 {% from "baseframe/bootstrap3/forms.html.jinja2" import renderform, ajaxform, widget_ext_scripts, widgetscripts %}
 {{ renderform(form=form, formid=formid, ref_id=ref_id, submit=submit, message=message, cancel_url=cancel_url, multipart=multipart) }}
-{{ ajaxform(ref_id=ref_id, request=request, force=true) }}
+{%- if with_chrome -%}
+  {{ ajaxform(ref_id=ref_id, request=request, force=true) }}
+{%- endif -%}
```
jace: This appears to force ajaxform on all forms even if the app did not want it. This may break non-Ajax forms in some apps. Confirm, please?

PR #185, opened 2018-02-28
Added EnumField and tests

### LINKED ISSUE #184: EnumField required
Baseframe needs an `EnumField` that builds on `SelectField` to offer support for `LabeledEnum`. Related: hasgeek/coaster#173

The problem with `SelectField` is that it runs the coerce operation on raw form data, _before_ validating against choices. To convert from `LabledEnum`'s name to value (using `.value_for()`), we need to do it _after_ validation. The `EnumField` therefore must do two things on top of `SelectField`:

1. Accept a `LabeledEnum` for the `choices` parameter and retrieve choices by calling `.nametitles()`
2. After validation, set `self.data = self.lenum.value_for(self.data)`

The latter can be achieved with a custom `process_formdata` method:

```python
class EnumField(SelectField):
    …
    def process_formdata(self, valuelist):
        retval = super(EnumField, self).process_formdata(valuelist)
        if self.data is not None:
            self.data = self.lenum.value_for(self.data)
        return retval
```

iambibhas:
Fixes #184.

### DIFF
--- a/baseframe/forms/fields.py
+++ b/baseframe/forms/fields.py
@@ -5,6 +5,7 @@ from decimal import Decimal, InvalidOperation as DecimalError
 from six.moves.urllib.parse import urljoin
 from pytz import utc, timezone as pytz_timezone
 from flask import current_app
+from coaster.utils import LabeledEnum
 import wtforms
 from wtforms.fields import SelectField as SelectFieldBase, SelectMultipleField, SubmitField, FileField
 from wtforms.compat import text_type
@@ -18,7 +19,7 @@ from .parsleyjs import TextAreaField, StringField, URLField
 
 __all__ = ['SANITIZE_TAGS', 'SANITIZE_ATTRIBUTES',
     'TinyMce3Field', 'TinyMce4Field', 'RichTextField', 'DateTimeField', 'TextListField',
-    'AnnotatedTextField', 'MarkdownField', 'StylesheetField', 'ImgeeField',
+    'AnnotatedTextField', 'MarkdownField', 'StylesheetField', 'ImgeeField', 'EnumField',
     'FormField', 'UserSelectField', 'UserSelectMultiField', 'GeonameSelectField', 'GeonameSelectMultiField',
     'CoordinatesField', 'RadioMatrixField', 'AutocompleteField', 'AutocompleteMultipleField', 'SelectField',
     # Imported from WTForms:
@@ -673,3 +674,15 @@ class RadioMatrixField(wtforms.Field):
         for fname, ftitle in self.fields:
             if fname in self.data:
                 setattr(obj, fname, self.data[fname])
+
+
+class EnumField(SelectField):
+    def __init__(self, *args, **kwargs):
+        super(EnumField, self).__init__(*args, **kwargs)
+        self.lenum = self.choices
+        self.choices = self.choices.nametitles()
+
+    def post_validate(self, form, validation_stopped):
+        super(EnumField, self).post_validate(form, validation_stopped)
+        if self.data is not None:
+            self.data = self.lenum.value_for(self.data)
--- a/baseframe/forms/fields.py
+++ b/baseframe/forms/fields.py
@@ -687,12 +687,13 @@ class EnumSelectField(SelectField):
 
     """
     def __init__(self, *args, **kwargs):
-        default = kwargs.pop('default')
+        default = kwargs.pop('default', None)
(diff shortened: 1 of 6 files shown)


baseframe/forms/form.py
```diff
@@ -38,6 +38,7 @@
     'AnnotatedTextField': bfields.AnnotatedTextField,
     'MarkdownField': bfields.MarkdownField,
     'ImageField': bfields.ImgeeField,
+    'EnumField': bfields.EnumField,
```
jace: No EnumField here. This is for the form generator, which does not have access to enums.

tests/test_fields.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class MY_LENUM(LabeledEnum):
+    FIRST = (1, 'first', __("First"))
+    SECOND = (2, 'second', __("Second"))
+    THIRD = (3, 'third', __("Third"))
+
+
+class MyForm(forms.Form):
+    position = forms.EnumField(__("Position"), choices=MY_LENUM, default=MY_LENUM.THIRD)
```
jace: Your test is missing a test for the `default`. This particular usage will fail, since there’s no translation from value to name.

iambibhas: Can we do this? 
```
class EnumField(SelectField):
    def __init__(self, *args, **kwargs):
        super(EnumField, self).__init__(*args, **kwargs)
        self.lenum = self.choices
        self.choices = self.choices.nametitles()
        if self.default:
            self.default = self.lenum.get(self.default).name
```

jace: Let's do this:

1. Rename the field to `LabeledEnumField`
2. Change the parameter to `lenum`. If `lenum` is provided, `self.choices` is set from it.
3. `default = kwargs.pop('default')` (before calling `super`, followed by) `self.default = self.lenum[default].name`

jace: Using `lenum` instead of `choices` is analogous to how `QuerySelectField` accepts `query` instead of `choices`.

jace: Actually, strike `LabeledEnumField`. For consistency this should be `EnumSelectField`.

baseframe/forms/fields.py
```diff
@@ -673,3 +674,27 @@ def populate_obj(self, obj, name):
...
+    Take a LabeledEnum and set choices based on that,
+    also doesn't reveal the actual value through the form
+
+    Takes a `lenum` argument instead of `choices`. E.g. -
+
+        position = forms.EnumSelectField(__("Position"), lenum=MY_LENUM, default=MY_LENUM.THIRD)
+
+    """
+    def __init__(self, *args, **kwargs):
+        default = kwargs.pop('default')
```
jace: This will fail if there was no `default` in `kwargs`.

baseframe/forms/fields.py
```diff
@@ -673,3 +674,27 @@ def populate_obj(self, obj, name):
...
+        self.lenum = kwargs.pop('lenum')
+
+        super(EnumSelectField, self).__init__(*args, **kwargs)
+        self.choices = self.lenum.nametitles()
+        self.default = self.lenum[default].name
```
jace: This also assumes default is specified.

baseframe/forms/fields.py
```diff
@@ -673,3 +674,28 @@ def populate_obj(self, obj, name):
...
+
+
+class EnumSelectField(SelectField):
+    """
+    Take a LabeledEnum and set choices based on that,
+    also doesn't reveal the actual value through the form
+
+    Takes a `lenum` argument instead of `choices`. E.g. -
+
+        position = forms.EnumSelectField(__("Position"), lenum=MY_LENUM, default=MY_LENUM.THIRD)
```
jace: This docstring is not valid ReStructuredText.

tests/test_fields.py
```diff
@@ -0,0 +1,41 @@
...
+        baseframe.init_app(app, requires=['baseframe'])
+        self.ctx = app.test_request_context()
+        self.ctx.push()
+        self.form = MyForm(meta={'csrf': False})
+
+    def tearDown(self):
+        self.ctx.pop()
+
+    def test_enumfield(self):
+        self.assertEqual(self.form.position.data, 'third')  # because nor yet validated
```
jace: This feels wrong. Regardless of validation it should have the correct value. Look at how `QuerySelectField` does it. We're okay if they have the same problem.

PR #186, opened 2018-03-05
Added validators for webmail domain using mxsniff

### LINKED ISSUE #170: Webmail domains list should be deprecated
An analysis of Baseframe's `webmail_domains` list using [mxsniff](/jace/mxsniff) today showed that the list was wildly inaccurate, featuring many domains with no MX records at all (meaning they're very likely not email services at all), and some using Google G Suite and other commercial mail hosting services.

Since some webmail services like Mailinator insist on keeping their list of domains secret, there is a low chance the webmail domain list will ever be good enough. However, we can get better accuracy by compiling a list of MX servers for public email services, instead. mxsniff now includes this feature in release 0.3.1 (jace/mxsniff#4), so we should deprecate Baseframe's list, provide a helper function to wrap mxsniff's scanner, and switch over in all apps that use it (notably Hasjob).

iambibhas:
Fixes #170.

### DIFF
--- a/baseframe/forms/validators.py
+++ b/baseframe/forms/validators.py
@@ -164,7 +164,7 @@ class NotEqualTo(_Comparison):
 
 class IsPublicEmailDomain(object):
     """
-    Validate that field.data belongs to a public webmail domain.
+    Validate that field.data belongs to a public email domain.
     If the domain lookup fails and mxsniff raises ``MXLookupException``,
     this validator will fail.
 
@@ -215,7 +215,7 @@ class IsPublicEmailDomain(object):
 
 class IsNotPublicEmailDomain(IsPublicEmailDomain):
     """
-    Validate that field.data does not belong to a public webmail domain.
+    Validate that field.data does not belong to a public email domain.
     If the domain lookup fails and mxsniff raises ``MXLookupException``, this validator
     will still pass, as we expect that most domains are not public email domains.
 
@@ -232,7 +232,7 @@ class IsNotPublicEmailDomain(IsPublicEmailDomain):
         if sniffedmx is None or not any([p['public'] for p in sniffedmx['providers']]):
             # sniffedmx is None only if the domain lookup fails.
             # This validator will pass in that case because we assume
-            # that most domains are not public email domain.
+            # that most domains are not public email domains.
             return
         else:
             raise ValidationError(self.message)
--- a/baseframe/forms/validators.py
+++ b/baseframe/forms/validators.py
@@ -175,16 +175,13 @@ class IsPublicEmailDomain(object):
         self.message = message or _(u'This domain is not a public email domain.')
 
     def get_mx(self, email_or_domain):
-        def make_cache_key(email_or_domain):
-            if six.PY2:
-                cache_key = 'mxrecord/' + urlquote(
-                    email_or_domain.encode('utf-8') if isinstance(email_or_domain, six.text_type) else email_or_domain,
-                    safe='')
-            else:
-                cache_key = 'mxrecord/' + urlquote(email_or_domain, safe='')
-            return cache_key
+        if six.PY2:
+            cache_key = 'mxrecord/' + urlquote(
+                email_or_domain.encode('utf-8') if isinstance(email_or_domain, six.text_type) else email_or_domain,
+                safe='')
+        else:
+            cache_key = 'mxrecord/' + urlquote(email_or_domain, safe='')
 
-        cache_key = make_cache_key(email_or_domain)
         mx_cache = asset_cache.get(cache_key)
 
         if mx_cache and isinstance(mx_cache, dict):
--- a/baseframe/forms/validators.py
+++ b/baseframe/forms/validators.py
@@ -165,7 +165,8 @@ class NotEqualTo(_Comparison):
 class IsPublicEmailDomain(object):
     """
     Validate that field.data belongs to a public webmail domain.
-    If the domain raises mxsniff's ``MXLookupException``, this validator will fail.
+    If the domain does not exist and raises mxsniff's ``MXLookupException``,
+    this validator will fail.
 
     :param message:
         Error message to raise in case of a validation error.
--- a/baseframe/forms/validators.py
+++ b/baseframe/forms/validators.py
@@ -11,7 +11,7 @@ from wtforms.validators import (DataRequired, InputRequired, Optional, Length, E
 import requests
 from lxml import html
 from coaster.utils import make_name, deobfuscate_email
-from mxsniff import mxsniff
+from mxsniff import mxsniff, MXLookupException
 from .. import b_ as _, b__ as __, asset_cache
 from ..signals import exception_catchall
 
@@ -163,14 +163,27 @@ class NotEqualTo(_Comparison):
 
 
 class IsPublicEmailDomain(object):
+    """
+    Validate that field.data belongs to a public webmail domain.
+    If the domain raises mxsniff's ``MXLookupException``, this validator will fail.
+
+    :param message:
+        Error message to raise in case of a validation error.
+    """
     def __init__(self, message=None):
         self.message = message or _(u'This domain is not a public email domain.')
 
-    def make_cache_key(self, email_or_domain):
-        return 'mxrecord/' + email_or_domain
-
     def get_mx(self, email_or_domain):
-        cache_key = self.make_cache_key(email_or_domain)
+        def make_cache_key(email_or_domain):
+            if six.PY2:
+                cache_key = 'mxrecord/' + urlquote(
+                    email_or_domain.encode('utf-8') if isinstance(email_or_domain, six.text_type) else email_or_domain,
+                    safe='')
+            else:
+                cache_key = 'mxrecord/' + urlquote(email_or_domain, safe='')
+            return cache_key
+
+        cache_key = make_cache_key(email_or_domain)
         mx_cache = asset_cache.get(cache_key)
 
         if mx_cache and isinstance(mx_cache, dict):
@@ -179,9 +192,13 @@ class IsPublicEmailDomain(object):
             domain = None
 
         if domain is None:
-            # Something is wrong with the cache, fetch result again and set cache
-            sniffedmx = mxsniff(email_or_domain)
-            asset_cache.set(cache_key, sniffedmx)
+            # Cache entry missing or corrupted; fetch a new result and update cache
+            try:
+                sniffedmx = mxsniff(email_or_domain)
+            except MXLookupException:
+                # Domain does not exist
+                return
+            asset_cache.set(cache_key, sniffedmx, timeout=86400)
             return sniffedmx
         else:
             # cache is fine, return it
@@ -189,19 +206,28 @@ class IsPublicEmailDomain(object):
 
     def __call__(self, form, field):
         sniffedmx = self.get_mx(field.data)
-        if any([p['public'] for p in sniffedmx['providers']]):
+        if sniffedmx is not None and any([p['public'] for p in sniffedmx['providers']]):
             return
         else:
             raise ValidationError(self.message.format(domain=field.data))
 
 
 class IsNotPublicEmailDomain(IsPublicEmailDomain):
+    """
+    Validate that field.data does not belong to a public webmail domain.
+    If the domain does not exist and raises mxsniff's ``MXLookupException``, this validator
+    will still pass, as we expect that most domains are not public email domains.
+
+    :param message:
+        Error message to raise in case of a validation error.
+    """
+
     def __init__(self, message=None):
         self.message = message or _(u'This domain is a public email domain.')
 
     def __call__(self, form, field):
         sniffedmx = self.get_mx(field.data)
-        if not any([p['public'] for p in sniffedmx['providers']]):
+        if sniffedmx is None or not any([p['public'] for p in sniffedmx['providers']]):
             return
         else:
             raise ValidationError(self.message.format(domain=field.data))
--- a/baseframe/forms/validators.py
+++ b/baseframe/forms/validators.py
@@ -165,14 +165,15 @@ class NotEqualTo(_Comparison):
 class IsPublicEmailDomain(object):
     """
     Validate that field.data belongs to a public webmail domain.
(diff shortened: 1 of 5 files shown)


baseframe/forms/validators.py
```diff
@@ -161,6 +162,30 @@ def compare(self, value, other):
...
 
 
+class PublicEmailDomain(object):
+    def __init__(self, message=None):
+        self.message = message or _(u'The domain "{domain}" is not a public email domain.')
+
+    def __call__(self, form, field):
+        sniffedmx = mxsniff(field.data)
+        if len(sniffedmx['providers']) > 0:
+            return
```
jace: You need to check if any of the provider dictionaries has a key `'public'` with value `True`. The presence of a provider tells you nothing.

baseframe/forms/validators.py
```diff
@@ -161,6 +162,30 @@ def compare(self, value, other):
         return value != other
 
 
+class PublicEmailDomain(object):
+    def __init__(self, message=None):
+        self.message = message or _(u'The domain "{domain}" is not a public email domain.')
+
+    def __call__(self, form, field):
+        sniffedmx = mxsniff(field.data)
```
jace: I think the flag to look up the domain cache is False by default. If so, make it True.

baseframe/forms/validators.py
```diff
@@ -161,6 +162,30 @@ def compare(self, value, other):
...
+            raise StopValidation(self.message.format(domain=field.data))
+
+
+class NotPublicEmailDomain(object):
+    def __init__(self, message=None):
+        self.message = message or _(u'The domain "{domain}" is a public email domain.')
+
+    def __call__(self, form, field):
+        sniffedmx = mxsniff(field.data)
+        if len(sniffedmx['providers']) == 0:
```
jace: As above.

tests/test_validators.py
```diff
@@ -31,6 +31,20 @@ def test_invalid_url(self):
             self.form.process(url=url)
             self.assertEqual(self.form.validate(), False)
 
+    def test_public_email_domain(self):
+        with self.app.test_request_context('/'):
+            self.webmail_form.process(
+                webmail_domain='google.com',
```
jace: This is not a webmail domain. `gmail.com` is.

baseframe/forms/validators.py
iambibhas: Can I reuse the above validator inside this one?

baseframe/forms/validators.py
iambibhas: `cache` seems to be expecting a dict? Is the client app supposed to maintain the cache list? https://github.com/jace/mxsniff/blob/master/mxsniff/__init__.py#L185

baseframe/forms/validators.py
```diff
@@ -161,6 +162,30 @@ def compare(self, value, other):
...
+        else:
+            raise StopValidation(self.message.format(domain=field.data))
+
+
+class NotPublicEmailDomain(object):
+    def __init__(self, message=None):
+        self.message = message or _(u'The domain "{domain}" is a public email domain.')
+
+    def __call__(self, form, field):
+        sniffedmx = mxsniff(field.data, cache=True)
```
jace: As above.

baseframe/forms/validators.py
```diff
@@ -161,6 +162,30 @@ def compare(self, value, other):
...
+class PublicEmailDomain(object):
+    def __init__(self, message=None):
+        self.message = message or _(u'The domain "{domain}" is not a public email domain.')
+
+    def __call__(self, form, field):
+        sniffedmx = mxsniff(field.data, cache=True)
+        if any([p['public'] for p in sniffedmx['providers']]):
+            return
+        else:
+            raise StopValidation(self.message.format(domain=field.data))
```
jace: `StopValidation` prevents other validators from running. Use `ValidationError` instead.

baseframe/forms/validators.py
```diff
@@ -161,6 +162,30 @@ def compare(self, value, other):
...
+class NotPublicEmailDomain(object):
+    def __init__(self, message=None):
+        self.message = message or _(u'The domain "{domain}" is a public email domain.')
+
+    def __call__(self, form, field):
+        sniffedmx = mxsniff(field.data, cache=True)
+        if not any([p['public'] for p in sniffedmx['providers']]):
+            return
+        else:
+            raise StopValidation(self.message.format(domain=field.data))
```
jace: As above.

baseframe/forms/validators.py
```diff
@@ -161,6 +162,30 @@ def compare(self, value, other):
         return value != other
 
 
+class PublicEmailDomain(object):
```
jace: I think we should make this `IsPublicEmailDomain` since the name is slightly ambiguous.

baseframe/forms/validators.py
```diff
@@ -161,6 +162,30 @@ def compare(self, value, other):
...
+    def __init__(self, message=None):
+        self.message = message or _(u'The domain "{domain}" is not a public email domain.')
+
+    def __call__(self, form, field):
+        sniffedmx = mxsniff(field.data, cache=True)
```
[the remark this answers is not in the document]
iambibhas: @jace looks like mxsniff can't take an empty dict for the `cache` param. is that intentional?

baseframe/forms/validators.py
```diff
@@ -161,6 +162,67 @@ def compare(self, value, other):
...
+        cache_key = self.make_cache_key(key)
+        asset_cache.set(cache_key, val, timeout=86400)
+
+    def __contains__(self, key):
+        cache_key = self.make_cache_key(key)
+        mx_cache = asset_cache.get(cache_key)
+        return mx_cache is not None
+
+
+_mxsniff_cache = MxSniffCache()
```
jace: Oh, no, no, no. You can't do process-globals in a multi-threaded app, and it's completely pointless in a multi-process deployment. Use Redis. Look at how the `ValidEmail` validator does it.

baseframe/forms/validators.py
```diff
@@ -161,6 +162,67 @@ def compare(self, value, other):
...
+
+class IsPublicEmailDomain(object):
+    def __init__(self, message=None):
+        self.message = message or _(u'The domain "{domain}" is not a public email domain.')
```
jace: Don't play back user input in an error message. That's asking to be embarrassed. Change to "This domain is not a public email domain."

baseframe/forms/validators.py
```diff
@@ -161,6 +162,67 @@ def compare(self, value, other):
...
+        sniffedmx = _mxsniff_cache[field.data]
+        if any([p['public'] for p in sniffedmx['providers']]):
+            return
+        else:
+            raise ValidationError(self.message.format(domain=field.data))
+
+
+class IsNotPublicEmailDomain(object):
+    def __init__(self, message=None):
+        self.message = message or _(u'The domain "{domain}" is a public email domain.')
```
jace: As above.

baseframe/forms/validators.py
```diff
@@ -161,6 +162,51 @@ def compare(self, value, other):
...
+        cache_key = self.make_cache_key(email_or_domain)
+        mx_cache = asset_cache.get(cache_key)
+
+        if mx_cache and isinstance(mx_cache, dict):
+            domain = mx_cache.get('domain')
+        else:
+            domain = None
+
+        if domain is None:
+            # Something is wrong with the cache, fetch result again and set cache
```
jace: Change comment to `Cache entry missing or corrupted; fetch a new result and update cache`

baseframe/forms/validators.py
```diff
@@ -161,6 +162,51 @@ def compare(self, value, other):
         return value != other
 
 
+class IsPublicEmailDomain(object):
+    def __init__(self, message=None):
+        self.message = message or _(u'This domain is not a public email domain.')
+
+    def make_cache_key(self, email_or_domain):
+        return 'mxrecord/' + email_or_domain
```
jace: This function can be moved inline into `get_mx` as it's only called once.

jace: You may need the `six.PY2` check as used in `ValidUrl`. Add an Emoji domain to your tests to confirm the cache key works. Pick something from https://i❤.ws/ (https://xn--i-7iq.ws/)

baseframe/forms/validators.py
```diff
@@ -161,6 +162,51 @@ def compare(self, value, other):
...
+        mx_cache = asset_cache.get(cache_key)
+
+        if mx_cache and isinstance(mx_cache, dict):
+            domain = mx_cache.get('domain')
+        else:
+            domain = None
+
+        if domain is None:
+            # Something is wrong with the cache, fetch result again and set cache
+            sniffedmx = mxsniff(email_or_domain)
```
jace: Check for exceptions here. `mxsniff`'s documentation includes this:

```rst
:raises MXLookupException: If a DNS lookup error happens and ``ignore_errors`` is False
```

If an exception happens, the validator must:

1. Fail in `IsPublicEmailDomain` (since we were unable to confirm that it _is_ a public email domain)
2. Pass in `IsNotPublicEmailDomain` (since our expectation is that most domains are not public email domains)

The documentation for both validators must clearly specify this behaviour.

baseframe/forms/validators.py
```diff
@@ -161,6 +162,51 @@ def compare(self, value, other):
...
+            asset_cache.set(cache_key, sniffedmx)
```
jace: What's the cache timeout? This can't be indefinite. Use 24 hours (86,400 seconds).

baseframe/forms/validators.py
```diff
@@ -161,6 +162,75 @@ def compare(self, value, other):
         return value != other
 
 
+class IsPublicEmailDomain(object):
+    """
+    Validate that field.data belongs to a public webmail domain.
+    If the domain does not exist and raises mxsniff's ``MXLookupException``,
```
jace: Domain not existing will not raise that exception. A timeout will.

baseframe/forms/validators.py
```diff
@@ -161,6 +162,75 @@ def compare(self, value, other):
...
+            domain = mx_cache.get('domain')
+        else:
+            domain = None
+
+        if domain is None:
+            # Cache entry missing or corrupted; fetch a new result and update cache
+            try:
+                sniffedmx = mxsniff(email_or_domain)
+            except MXLookupException:
+                # Domain does not exist
```
jace: Incorrect. Lookup failed (timeout, no DNS server configured, etc).

baseframe/forms/validators.py
```diff
@@ -161,6 +162,75 @@ def compare(self, value, other):
...
+    :param message:
+        Error message to raise in case of a validation error.
+    """
+
+    def __init__(self, message=None):
+        self.message = message or _(u'This domain is a public email domain.')
+
+    def __call__(self, form, field):
+        sniffedmx = self.get_mx(field.data)
+        if sniffedmx is None or not any([p['public'] for p in sniffedmx['providers']]):
```
jace: Add a comment explaining the None value.

tests/test_validators.py
```diff
@@ -31,6 +32,50 @@ def test_invalid_url(self):
...
+            self.webmail_form.process(
+                webmail_domain='gmail.com',
+                not_webmail_domain='foobar.com'
+            )
+            self.assertTrue(self.webmail_form.validate())
+
+            # Emoji domain, should not raise any exception
+            self.webmail_form.process(
+                webmail_domain='gmail.com',
+                not_webmail_domain='i❤.ws'
```
jace: How is this working if not declared as a Unicode string? You don’t have Unicode literals enabled.

tests/test_validators.py
```diff
@@ -31,6 +32,50 @@ def test_invalid_url(self):
...
+            # these domain names don't exist, MX lookup will fail
+            # So, webmail_domain should fail, but not_webmail_domain should pass
+            self.webmail_form.process(
+                webmail_domain='9vGw1TMChQWH.com',
+                not_webmail_domain='9vGw1TMChQWH.com'
+            )
+            self.assertFalse(self.webmail_form.validate())
+            self.assertIn('webmail_domain', self.webmail_form.errors)
+            self.assertNotIn('not_webmail_domain', self.webmail_form.errors)
+
```
jace: Please pass the same values to both validators in all tests, to confirm that they always exhibit opposite responses.

jace: Please test the MXLookupException handler as well. Accept a timeout parameter in both validators, and in one test change the timeout to a very small number.

baseframe/forms/validators.py
```diff
@@ -161,6 +162,82 @@ def compare(self, value, other):
         return value != other
 
 
+class IsPublicEmailDomain(object):
+    """
+    Validate that field.data belongs to a public webmail domain.
```
jace: `public email domain` (not `webmail`)

baseframe/forms/validators.py
```diff
@@ -161,6 +162,82 @@ def compare(self, value, other):
...
+            return
+        else:
+            # sniffedmx is None only if the domain lookup fails.
+            # This validator will fail in that case
+            raise ValidationError(self.message.format(domain=field.data))
+
+
+class IsNotPublicEmailDomain(IsPublicEmailDomain):
+    """
+    Validate that field.data does not belong to a public webmail domain.
```
jace: `public email domain` again

baseframe/forms/validators.py
```diff
@@ -161,6 +162,82 @@ def compare(self, value, other):
...
+    def __init__(self, message=None, timeout=30):
+        self.message = message or _(u'This domain is a public email domain.')
+        self.timeout = timeout
+
+    def __call__(self, form, field):
+        sniffedmx = self.get_mx(field.data, timeout=self.timeout)
+        if sniffedmx is None or not any([p['public'] for p in sniffedmx['providers']]):
+            # sniffedmx is None only if the domain lookup fails.
+            # This validator will pass in that case because we assume
+            # that most domains are not public email domain.
```
jace: `public email domains` (plural)

jace:
Looks like we have everything _except_ the `MXLookupError` test.

PR #187, opened 2018-03-20
Added helper function for public email domain check

iambibhas:
Brings us closer to #170.


jace:
You need to do this the other way around: make the function standalone, and make the validator depend on the function.

baseframe/forms/helper.py
```diff
@@ -0,0 +1,55 @@
...
+from six.moves.urllib.parse import quote as urlquote
+from .. import asset_cache
+from mxsniff import mxsniff, MXLookupException
+
+__all__ = ['is_public_email_domain']
+
+
+def is_public_email_domain(email_or_domain, timeout=30):
+    """
+    Checks if the given email address or domain belongs to a public email domain
```
jace: "Checks if the given domain (or domain of given email) is known to offer public email accounts"

baseframe/forms/helper.py
```diff
@@ -0,0 +1,55 @@
...
+
+    :param email_or_domain: email address or domain name to check
+    :param timeout: seconds after which mxsniff will give up retrieving MX data
```
jace: "Lookup timeout (will raise an exception if timeout happens before the check is completed and no default is provided)"

baseframe/forms/helper.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+    def get_mx(email_or_domain, timeout):
```
jace: Why is there a nested function? Move it to top level.

baseframe/forms/helper.py
```diff
@@ -0,0 +1,55 @@
+import six
+from six.moves.urllib.parse import quote as urlquote
+from .. import asset_cache
+from mxsniff import mxsniff, MXLookupException
+
+__all__ = ['is_public_email_domain']
+
+
+def is_public_email_domain(email_or_domain, timeout=30):
```
jace: Accept a `default=None` parameter.

baseframe/forms/helper.py
```diff
@@ -0,0 +1,55 @@
...
+        if mx_cache and isinstance(mx_cache, dict):
+            domain = mx_cache.get('domain')
+        else:
+            domain = None
+
+        if domain is None:
+            # Cache entry missing or corrupted; fetch a new result and update cache
+            try:
+                sniffedmx = mxsniff(email_or_domain, timeout=timeout)
+            except MXLookupException:
```
jace: If no default is provided, re-raise the exception.

baseframe/forms/helper.py
```diff
@@ -0,0 +1,55 @@
...
+            # cache is fine, return it
+            return mx_cache
+
+    sniffedmx = get_mx(email_or_domain, timeout=timeout)
+    if sniffedmx is not None and any([p['public'] for p in sniffedmx['providers']]):
+        return True
+    else:
+        # sniffedmx is None only if the domain lookup fails,
+        # in that case we return False because we assume that most
+        # domains are not public email domains
```
jace: Don't assume, return caller's default.

baseframe/forms/validators.py
```diff
@@ -175,45 +175,14 @@ def __init__(self, message=None, timeout=30):
...
-            asset_cache.set(cache_key, sniffedmx, timeout=86400)
-            return sniffedmx
-        else:
-            # cache is fine, return it
-            return mx_cache
-
     def __call__(self, form, field):
-        sniffedmx = self.get_mx(field.data, timeout=self.timeout)
-        if sniffedmx is not None and any([p['public'] for p in sniffedmx['providers']]):
+        if is_public_email_domain(field.data, timeout=self.timeout):
```
jace: Supply default here

baseframe/forms/validators.py
```diff
@@ -228,11 +197,9 @@ def __init__(self, message=None, timeout=30):
         self.timeout = timeout
 
     def __call__(self, form, field):
-        sniffedmx = self.get_mx(field.data, timeout=self.timeout)
-        if sniffedmx is None or not any([p['public'] for p in sniffedmx['providers']]):
-            # sniffedmx is None only if the domain lookup fails.
-            # This validator will pass in that case because we assume
-            # that most domains are not public email domains.
+        if not is_public_email_domain(field.data, timeout=self.timeout):
```
jace: Supply default here

tests/test_validators.py
```diff
@@ -72,6 +73,15 @@ def test_public_email_domain(self):
             self.assertIn('webmail_domain', self.webmail_form.errors)
             self.assertNotIn('not_webmail_domain', self.webmail_form.errors)
 
+    def test_public_email_domain_helper(self):
+        with self.app.test_request_context('/'):
+            self.assertEqual(is_public_email_domain(u'gmail.com'), True)
+            self.assertEqual(is_public_email_domain(u'i❤.ws'), False)
```
jace: Use a better known domain than i❤.ws for your tests. That domain is only for the IDN test. You could use contrasts such as `gmail.com` vs `google.com`.

baseframe/forms/validators.py
```diff
@@ -228,11 +197,7 @@ def __init__(self, message=None, timeout=30):
         self.timeout = timeout
 
     def __call__(self, form, field):
-        sniffedmx = self.get_mx(field.data, timeout=self.timeout)
-        if sniffedmx is None or not any([p['public'] for p in sniffedmx['providers']]):
-            # sniffedmx is None only if the domain lookup fails.
-            # This validator will pass in that case because we assume
-            # that most domains are not public email domains.
+        if not is_public_email_domain(field.data, default=False, timeout=self.timeout):
```
jace: Are your defaults correct? I get the feeling one of the defaults is wrong.

baseframe/utils.py
```diff
@@ -0,0 +1,59 @@
+import six
```
jace: Needs `# -*- coding: utf-8 -*-` on first line.

jace: Also needs `from __future__ import absolute_import`

baseframe/utils.py
```diff
@@ -0,0 +1,59 @@
...
+from six.moves.urllib.parse import quote as urlquote
+from . import asset_cache
+from mxsniff import mxsniff, MXLookupException
```
jace: `mxsniff` is an external dependency and must be imported after `six` and before `.`

baseframe/utils.py
```diff
@@ -0,0 +1,59 @@
...
+    :param default: default value to return in case domain lookup fails
+    :param timeout: Lookup timeout (will raise an exception if timeout happens before the check is completed and no default is provided)
+    """
+    sniffedmx = get_mx(email_or_domain, default=default, timeout=timeout)
+    if sniffedmx is not None and any([p['public'] for p in sniffedmx['providers']]):
+        return True
+    else:
+        # sniffedmx is only None if the domain lookup failed and a default is provided,
+        # in that case return default
+        return default
```
jace: Why are these two separate functions? If there is no reason to call `get_mx` separately from `is_public_email_domain`, make it inline.

tests/test_validators.py
```diff
@@ -72,6 +74,22 @@ def test_public_email_domain(self):
...
+            self.assertEqual(is_public_email_domain(u'gmail.com', default=False), True)
+            self.assertEqual(is_public_email_domain(u'google.com', default=False), False)
+            # this domain lookup fails, so the helper should return default value
+            self.assertEqual(
+                is_public_email_domain(u'www.abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijks.com', default=False),
+                False)
+            self.assertEqual(
+                is_public_email_domain(u'www.abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijks.com', default=True),
+                True)
+            # as no default is provided and the domain will trigger MXLookupExceptiom
```
jace: `and the domain lookup will trigger`

tests/test_validators.py
```diff
@@ -72,6 +74,22 @@ def test_public_email_domain(self):
...
+            # the helper function will raise as expection
+            with self.assertRaises(MXLookupException):
+                is_public_email_domain(u'www.abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijks.com')
```
jace: Place this test above the tests with the default, so that it's obvious what the default is doing there.

baseframe/forms/validators.py
iambibhas: Our assumption for validation is that in case of lookup failure, a domain is not a public email domain. so default is False for both cases. It's `not False` above in `IsNotPublicEmailDomain`, so the validation passes and `False` in `IsPublicEmailDomain`, so validation fails.

baseframe/utils.py
```diff
@@ -0,0 +1,47 @@
...
+__all__ = ['is_public_email_domain']
+
+
+def is_public_email_domain(email_or_domain, default=None, timeout=30):
+    """
+    Checks if the given domain (or domain of given email) is known to offer public email accounts
+
+    :param email_or_domain: email address or domain name to check
+    :param default: default value to return in case domain lookup fails
+    :param timeout: Lookup timeout (will raise an exception if timeout happens before the check is completed and no default is provided)
```
jace: The parenthesised bit can be moved into the main documentation. Also add `:raises MXLookupException: If a DNS lookup error happens and no default is specified`

baseframe/utils.py
```diff
@@ -0,0 +1,47 @@
...
+
+__all__ = ['is_public_email_domain']
+
+
+def is_public_email_domain(email_or_domain, default=None, timeout=30):
+    """
+    Checks if the given domain (or domain of given email) is known to offer public email accounts
+
+    :param email_or_domain: email address or domain name to check
+    :param default: default value to return in case domain lookup fails
```
jace: These two descriptions begin with a lowercase letter, while the next one begins with uppercase. Consistency?

baseframe/utils.py
```diff
@@ -0,0 +1,47 @@
...
+from mxsniff import mxsniff, MXLookupException
+
+from . import asset_cache
+
+__all__ = ['is_public_email_domain']
+
+
+def is_public_email_domain(email_or_domain, default=None, timeout=30):
+    """
+    Checks if the given domain (or domain of given email) is known to offer public email accounts
```
jace: Documentation should specify that Baseframe's cache is used.

baseframe/utils.py
```diff
@@ -0,0 +1,47 @@
...
+
+    if sniffedmx is None or not isinstance(sniffedmx, dict):
+        # Cache entry missing or corrupted; fetch a new result and update cache
+        try:
+            sniffedmx = mxsniff(email_or_domain, timeout=timeout)
+            asset_cache.set(cache_key, sniffedmx, timeout=86400)
+        except MXLookupException as e:
+            # Domain lookup failed
+            if default is None:
+                raise e
```
jace: `else: return default` here

baseframe/utils.py
```diff
@@ -0,0 +1,47 @@
...
+
+    if sniffedmx is None:
```
jace: What is the likelihood `mxsniff` returns `None`? If that's not documented behaviour, move default handling into the `except` block above, and assume you have a dictionary result here.

tests/test_validators.py
```diff
@@ -72,6 +74,23 @@ def test_public_email_domain(self):
             self.assertIn('webmail_domain', self.webmail_form.errors)
             self.assertNotIn('not_webmail_domain', self.webmail_form.errors)
 
+    def test_public_email_domain_helper(self):
+        with self.app.test_request_context('/'):
+            # As no default is provided and the domain lookup will trigger MXLookupExceptiom
+            # the helper function will raise as expection
```
jace: Comment could be better:

`Intentionally trigger a DNS lookup failure using an invalid domain name. Since no default is provided, we will receive an exception`

tests/test_validators.py
```diff
@@ -72,6 +74,23 @@ def test_public_email_domain(self):
...
+
+            self.assertEqual(is_public_email_domain(u'gmail.com', default=False), True)
+            self.assertEqual(is_public_email_domain(u'google.com', default=False), False)
+            # this domain lookup fails, so the helper should return default value
+            self.assertEqual(
+                is_public_email_domain(u'www.abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijks.com', default=False),
+                False)
+            self.assertEqual(
+                is_public_email_domain(u'www.abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijks.com', default=True),
+                True)
```
jace: These two should immediately follow the exception handler test.

tests/test_validators.py
```diff
@@ -72,6 +74,23 @@ def test_public_email_domain(self):
...
 
+    def test_public_email_domain_helper(self):
+        with self.app.test_request_context('/'):
+            # As no default is provided and the domain lookup will trigger MXLookupExceptiom
+            # the helper function will raise as expection
+            with self.assertRaises(MXLookupException):
+                is_public_email_domain(u'www.abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijks.com')
+
+            self.assertEqual(is_public_email_domain(u'gmail.com', default=False), True)
+            self.assertEqual(is_public_email_domain(u'google.com', default=False), False)
```
jace: Move these two to the top, since expected usage comes first.

PR #188, opened 2018-04-13
Fix autofocus bug

vidya-ram:
From Jinja 2.9, global scoped variables can't be modified inside a loop. https://github.com/pallets/jinja/issues/641. Hence using a dictionary to set autofocus and to check if it is first field of the form

Checks to set autofocus
Autofocus is set for the first field of the form provided it is not radio or checkbox.
In case of validation errors, autofocus is set for the first field of the form with an error.

### DIFF
--- a/baseframe/templates/baseframe/bootstrap3/forms.html.jinja2
+++ b/baseframe/templates/baseframe/bootstrap3/forms.html.jinja2
@@ -111,13 +111,15 @@
   {% if message %}<p>{{ message }}</p>{% endif %}
   <div style="display:none;"><input type="hidden" name="form.id" value="{{ formid }}" /></div>
   {{ form.hidden_tag() }}
-  {%- set autofocus = true %}
   {% for field in form -%}
     {%- if field.type in ['CSRFTokenField', 'HiddenField'] -%}
       {# Don't show hidden #}
     {%- else -%}
-      {{ renderfield(field, autofocus=autofocus, style=style) }}
-      {%- if autofocus %}{% set autofocus = false %}{% endif %}
+      {% if loop.first %}
+        {{ renderfield(field, autofocus=true, style=style) }}
+      {%- else -%}
+        {{ renderfield(field, style=style) }}
+      {% endif %}
     {%- endif %}
   {% endfor %}
 {%- endmacro %}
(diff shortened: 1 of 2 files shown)


baseframe/templates/baseframe/bootstrap3/forms.html.jinja2
```diff
@@ -111,13 +111,34 @@
   {% if message %}<p>{{ message }}</p>{% endif %}
   <div style="display:none;"><input type="hidden" name="form.id" value="{{ formid }}" /></div>
   {{ form.hidden_tag() }}
-  {%- set autofocus = true %}
+  <!--  In jinja, global scoped variables can't be modified inside a loop. https://github.com/pallets/jinja/issues/641. Hence using a dictionary to set autofocus and check if it is first field.
+  Autofocus is set true for the first field of the form provided it is not radio or checkbox.
+  In case of validation errors, autofocus is set true for the first field with an error.-->
```
jace: Use Jinja2 comments (`{# Comment #}`). HTML comments will be included in the rendered page.

jace: Also, this code seems a bit obtuse. Try this:

1. Figure out autofocus field (if any at all) in Python code
2. Pass the field object to `renderform`
3. Let renderform pass `autofocus=True` to `renderfield`

jace:
We should revisit this. It may be just 3-4 lines of Python code to identify the field to focus on:

1. If an error, the first non-hidden erroring field (in case of CSRF errors, no other field will have an error, so pass on to the next condition)
2. If no error, or no non-hidden error field, the first non-hidden field in the form
3. if no field is found (this form only contains hidden fields, such as with CSRF confirm forms, or no fields at all), no autofocus

Identify the field name, pass it on to `renderform`, and when `renderform` is rendering that field, it passes `autofocus=true`.

@iambibhas you should consider tackling this.

iambibhas:
@jace the deciding logic goes to the `FormGenerator` class?

jace:
@iambibhas No, `FormGenerator` is a JSON to Form convertor. It should go to the `render_form` function.

jace:
This PR addresses one of the accessibility concerns that Raghava brought up. @vidya-ram @iambibhas

PR #191, opened 2018-05-29
Add support to send custom dimension to GA.

vidya-ram:


baseframe/templates/baseframe/mui/baseframe.html.jinja2
```diff
@@ -150,6 +150,10 @@
       })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
 
       ga('create', {{ config['GA_CODE']|tojson }}, 'auto');
+      // Track Device Pixel Ratio
+      var pixelRatio = window.getDevicePixelRatio();
+      ga('set', 'metric1', pixelRatio);
+      {% block customgasettings -%}{%- endblock %}
```
jace: To avoid a blank line when there are no custom settings, you'll need `{% endblock -%}` (in both files).

jace: I'd also make the block name easier to read, like `custom_google_analytics` or `custom_ga_settings`. Whatever name we choose here, we'll be stuck with for years to come.

PR #194, opened 2018-08-16
Move static endpoint into a static subdomain

jace:
Instead of creating two static paths, use a single static subdomain to receive the benefits of browser caching.

This breaks Hasjob and other apps with external build systems that expect `/static` paths, so they need special handling.


jace:
Hasjob and Funnel have been upgraded to handle this change (in master branch). Other apps should not be affected. This is now safe to merge.

PR #198, opened 2018-09-20
Add Font-Awesome v4.7.0

vidya-ram:


baseframe/assets.py
```diff
@@ -203,6 +203,7 @@
 # Font Awesome
 assets['fontawesome.css'][Version('3.1.0')] = 'baseframe/css/fontawesome-3.1.css'
 assets['fontawesome.css'][Version('4.4.0')] = 'baseframe/css/fontawesome-4.4.css'
+assets['fontawesome.css'][Version('4.7.0')] = 'baseframe/css/fontawesome-4.7.css'
```
jace: We need to remove 4.4 when introducing 4.7, as there is no reason to use 4.4 anymore. 3.1 has been retained because the syntax is different and is still used by our older Bootstrap 2 websites (hacknight.in, etc).

PR #200, opened 2018-09-27
Move modernizr from head to body tag. Add appscripts block.

vidya-ram:
1. Move `Modernizr.js` from head to body tag, since scripts in head tag is rendering blocking
https://github.com/Modernizr/Modernizr/issues/878#issuecomment-41448059

2. Add `{% block appscripts %}{% endblock %}`
This block is to include JS assets of the app that are not required on all the pages but has to be included before baseframe bundle(assets "js_all").
For instance codemirror-markdown-js.js is only required in the formlayout pages. CodeMirror functions are called from baseframe.js, hence it has to be included before  baseframe.js

### DIFF
--- a/baseframe/templates/baseframe/bootstrap3/baseframe.html.jinja2
+++ b/baseframe/templates/baseframe/bootstrap3/baseframe.html.jinja2
@@ -46,7 +46,6 @@
     <link rel="stylesheet" type="text/css" href="{{ ASSET_URL }}" />
   {%- endassets %}
   <link rel="stylesheet" type="text/css" href="//fonts.googleapis.com/css?family=Source+Sans+Pro:400italic,600italic,400,600" />
-  <script src="{{ url_for('baseframe.static', filename='js/modernizr.min.js') }}"></script>
 
   {%- block typekit %}
     {%- if config['TYPEKIT_CODE'] %}
@@ -101,6 +100,7 @@
     {%- endblock %}
   {%- endblock %}
 
+  <script src="{{ url_for('baseframe.static', filename='js/modernizr.min.js') }}"></script>
   <!-- Included JS Files -->
   <!--[if lt IE 9]>
     <script src="//ajax.googleapis.com/ajax/libs/jquery/1.11.2/jquery.min.js"></script>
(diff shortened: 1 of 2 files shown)


baseframe/templates/baseframe/bootstrap3/baseframe.html.jinja2
```diff
@@ -121,6 +121,11 @@
   {%- for asset in config.get('ext_js', []) %}
     <script type="text/javascript" src="{{ asset|ext_asset_url }}"></script>
   {%- endfor %}
+
+  <!-- This block is to include JS assets of the app that are not required on all the pages but has to be included before baseframe bundle(assets "js_all").
```
jace: Use Jinja2 comments, not HTML comments. This should not render into the final page. Also use `{#- comment #}` (note the hyphen) so that an empty line is not inserted into the output.

baseframe/templates/baseframe/mui/baseframe.html.jinja2
```diff
@@ -138,6 +138,11 @@
   {%- for asset in config.get('ext_js', []) %}
     <script type="text/javascript" src="{{ asset|ext_asset_url }}"></script>
   {%- endfor %}
+
+  <!-- This block is to include JS assets of the app that are not required on all the pages but has to be included before baseframe bundle(assets "js_all").
```
jace: As above.

baseframe/templates/baseframe/bootstrap3/baseframe.html.jinja2
```diff
@@ -121,6 +121,11 @@
...
+  For instance codemirror-markdown-js.js is only required in the formlayout pages. CodeMirror functions are called from baseframe.js, hence it has to be included before  baseframe.js -->
+  {% block appscripts %}{% endblock %}
```
jace: Consider calling this `pagescripts` instead of `appscripts` as it's meant for specific pages.

baseframe/templates/baseframe/mui/baseframe.html.jinja2
```diff
@@ -138,6 +138,11 @@
...
+  For instance codemirror-markdown-js.js is only required in the formlayout pages. CodeMirror functions are called from baseframe.js, hence it has to be included before  baseframe.js -->
+  {% block appscripts %}{% endblock %}
```
jace: As above.

baseframe/templates/baseframe/bootstrap3/baseframe.html.jinja2
vidya-ram: Changed to `{#- comment #}`

PR #204, opened 2018-11-20
Add api key to google map api.

vidya-ram:


baseframe/templates/baseframe/bootstrap3/forms.html.jinja2
```diff
@@ -262,7 +262,8 @@
     {%- for field in form %}
       {%- if field.type == 'CoordinatesField' %}
         {#- We're using setdefault instead of a simple = because the do extension doesn't support assignment #}
-        {%- do scripts.setdefault('gmap', '//maps.google.com/maps/api/js?sensor=false&libraries=places') %}
+        {% set google_map_api = '//maps.google.com/maps/api/js?key=' + config["GOOGLE_MAPS_API_KEY"] + '&sensor=false&libraries=places' -%}
```
jace: We can make these mandatory `https:` from now on, so `https://maps.google.com/...`

jace: Use `config['GOOGLE_MAPS_API_KEY']` (single quotes, per code standard). This _will_ fail if the app hasn't set the API key, but that's okay because any app that uses a coordinates field needs to set it.

baseframe/templates/baseframe/bootstrap3/forms.html.jinja2
```diff
@@ -262,7 +262,8 @@
...
+        {%- do scripts.setdefault('gmap', google_map_api) %}
```
jace: Not sure why we're setting a variable and then storing it to this dict. Why not combine into one `do` statement?

baseframe/templates/baseframe/mui/forms.html.jinja2
```diff
@@ -280,7 +280,8 @@
     {%- for field in form %}
       {%- if field.type == 'CoordinatesField' %}
         {#- We're using setdefault instead of a simple = because the do extension doesn't support assignment #}
-        {%- do scripts.setdefault('gmap', '//maps.google.com/maps/api/js?sensor=false&libraries=places') %}
+        {% set google_map_api = '//maps.google.com/maps/api/js?key=' + config["GOOGLE_MAPS_API_KEY"] + '&sensor=false&libraries=places' -%}
+        {%- do scripts.setdefault('gmap', google_map_api) %}
```
jace: Likewise.

PR #205, opened 2018-11-28
Add leaflet for maps

vidya-ram:
1. Add jquery.succinct.js (jQuery plugin for truncating multiple lines of text)
2. Add leaflet.js for maps and leaflet-search plugin for search


baseframe/static/js/baseframe-material.js
```diff
@@ -370,6 +353,12 @@ window.Baseframe.MapMarker.prototype.getDefaultLocation = function() {
   return {latitude: latitude, longitude: longitude};
 };
 
+window.Baseframe.MapMarker.prototype.activate = function(){
+  var self = this;
```
jace: `self` is not being used in this function. What's it for?

vidya-ram: Removed

PR #206, opened 2019-01-11
Added JSONfield

iambibhas:
Relates to https://github.com/hasgeek/funnel/pull/340

### DIFF
--- a/baseframe/forms/fields.py
+++ b/baseframe/forms/fields.py
@@ -731,14 +731,17 @@ class EnumSelectField(SelectField):
 
 class JsonField(wtforms.TextAreaField):
     """
-    JsonField takes JSON string as user input. If no user input is provided as formdata,
-    ``default`` value is used if available. If ``default`` is not provided, it's set to ``None``.
+    A TextArea field that stores JSON string as Python dict/list.
 
-    - If float value is provided in user input, it's converted to ``Decimal`` object.
-    - Comments in JSON string are not allowed. Validation will fail if JSON string contains comments.
-    - If non-serializable objects are set manually to ``fieldname.data``, validation will fail.
-    - If ``prettyprint`` is set to ``True``, the JSON string will be formatted before
-    populating the field widget.
+    - If float value is provided in JSON string, it's converted to ``Decimal`` object.
+    - Comments in JSON string are not allowed.
+    - Non-serializable objects are not allowed when setting data explicitly to ``form.fieldname.data``.
+
+    Takes a ``prettyprint`` argument that is ``False`` by default. If ``True``, the JSON string will be
+    indented before populating the textarea.
+
+        class MyForm(forms.Form):
+            field = forms.JsonField(__("My Field"), prettyprint=True, default=DEFAULT_VALUE)
     """
     prettyprint_args = {'sort_keys': True, 'indent': 2}
 
--- a/baseframe/forms/fields.py
+++ b/baseframe/forms/fields.py
@@ -10,7 +10,6 @@ from wtforms.fields import SelectField as SelectFieldBase, SelectMultipleField, 
 from wtforms.widgets import Select as OriginalSelectWidget
 from wtforms.compat import text_type
 from wtforms.utils import unset_value
-from wtforms.validators import StopValidation
 import bleach
 import simplejson as json
 import six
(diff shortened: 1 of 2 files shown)


baseframe/forms/fields.py
```diff
@@ -725,3 +726,24 @@ def process_formdata(self, valuelist):
     def pre_validate(self, form):
         if self.data is _invalid_marker:
             raise ValueError(self.gettext('Not a valid choice'))
+
+
+class JSONField(wtforms.TextAreaField):
```
jace: Should be `JsonField`

baseframe/forms/fields.py
```diff
@@ -725,3 +726,24 @@ def process_formdata(self, valuelist):
...
+    def _value(self):
+        if self.raw_data:
+            return self.raw_data[0]
+        elif self.data:
+            return json.dumps(self.data, ensure_ascii=False)
+        else:
+            return json.dumps(self.default) if self.default else u"{}"
```
jace: Null values are acceptable, so don't fall back to `{}`

baseframe/forms/fields.py
```diff
@@ -725,3 +726,24 @@ def process_formdata(self, valuelist):
...
         if self.data is _invalid_marker:
             raise ValueError(self.gettext('Not a valid choice'))
+
+
+class JSONField(wtforms.TextAreaField):
+    def _value(self):
+        if self.raw_data:
+            return self.raw_data[0]
+        elif self.data:
+            return json.dumps(self.data, ensure_ascii=False)
```
jace: Are you sure you want `ensure_ascii=False`? https://stackoverflow.com/a/40421631/78903

baseframe/forms/fields.py
```diff
@@ -10,14 +10,16 @@
...
 import bleach
+import simplejson as json
 import six
 
 from .. import _, get_timezone
 from .widgets import TinyMce3, TinyMce4, DateTimeInput, CoordinatesInput, RadioMatrixInput, SelectWidget, Select2Widget
 from .parsleyjs import TextAreaField, StringField, URLField
 
-__all__ = ['SANITIZE_TAGS', 'SANITIZE_ATTRIBUTES',
+__all__ = ['SANITIZE_TAGS', 'SANITIZE_ATTRIBUTES', 'JsonField',
```
jace: The `__all__` list needs to be sorted according to one linter. Could you do that?

baseframe/forms/fields.py
```diff
@@ -725,3 +727,76 @@ def process_formdata(self, valuelist):
...
+        # when the form is rendered, and the likelihood of this field being
+        # used without being rendered is low.
+
+        self.data = value
+
+    def process_formdata(self, valuelist):
+        if valuelist:
+            value = valuelist[0]
+            if not value:
+                self.data = self.default if self.default else None
```
jace: Just occurred to me this will fail if `self.default == {}`. Many of our JSON fields are non-nullable with default value `{}`, so that should be preserved.

baseframe/forms/fields.py
```diff
@@ -725,3 +727,76 @@ def process_formdata(self, valuelist):
...
+            try:
+                data = json.loads(value, use_decimal=self.use_decimal)
+            except ValueError as e:
+                raise ValueError(_("Invalid JSON: {0!r}").format(e))
+            if self.require_dict:
+                if not isinstance(data, dict):
+                    raise ValueError(_("The JSON root must be a hash object"))
+            self.data = data
+
+    # def pre_validate(self, form):
```
jace: If this is not required, remove it.

tests/test_fields.py
```diff
@@ -16,20 +18,35 @@ class MY_ENUM(LabeledEnum):
     __order__ = (FIRST, SECOND, THIRD)
 
 
-class EnumSelectForm(forms.Form):
+DEFAULT_JSONDATA = {'key': u"val"}
```
jace: Add a test for default value `{}` (as indicated above).

baseframe/forms/fields.py
```diff
@@ -725,3 +728,66 @@ def process_formdata(self, valuelist):
...
+        Render the internal Python value as a JSON string.
+        Specialcase `None` to return an empty string instead of a JSON ``null``.
+        """
+        if self.raw_data:
+            # If we've received data from a form, render it as is. This allows
+            # invalid JSON to be presented back to the user for correction.
+            return self.raw_data[0]
+        elif self.data is not None:
+            return json.dumps(self.data, use_decimal=self.use_decimal, ensure_ascii=False, **self.prettyprint_args)
+        return u''
```
iambibhas: @jace missed a case here. When the db value is `{}` (for any model field with `default={}`), this returns `{}` to the html form instead of the form field default. You removed that part in your commit. Any reason why?

PR #208, opened 2019-02-06
Add flag for autosave support

vidya-ram:


baseframe/forms/auto.py
```diff
@@ -18,24 +18,28 @@ class ConfirmDeleteForm(Form):
     cancel = SubmitField(__(u"Cancel"))
 
 
-def render_form(form, title, message='', formid=None, submit=__(u"Submit"), cancel_url=None, ajax=False, with_chrome=True, autosave=False):
+def render_form(form, title, message='', formid=None, submit=__(u"Submit"), cancel_url=None, ajax=False, with_chrome=True, action='', autosave=False, last_revision_id=None):
     multipart = False
     ref_id = 'form-' + (formid or buid())
+    if not action:
+        action = request.base_url
```
jace: On second thoughts, we should be using `request.url`, because query parameters may have significance as well. http://flask.pocoo.org/docs/1.0/api/#flask.Request.url

baseframe/templates/baseframe/bootstrap3/forms.html.jinja2
```diff
@@ -111,6 +111,9 @@
   {% if message %}<p>{{ message }}</p>{% endif %}
   <div style="display:none;"><input type="hidden" name="form.id" value="{{ formid }}" /></div>
   {{ form.hidden_tag() }}
+  {%- if autosave %}
+    <input type="hidden" name="form.revision" id="revision" value="{{ last_revision_id if last_revision_id is not none }}" />
```
jace: It is unsafe to use `id="revision"` because there could be multiple forms on the page. This is where `formid` matters. It's safe to use `id="{{ formid }}-revision"`, or to skip the id entirely and use CSS selector `form#{{ formid }} > input[name="form.revision"]`

baseframe/forms/auto.py
```diff
@@ -18,24 +18,28 @@ class ConfirmDeleteForm(Form):
     cancel = SubmitField(__(u"Cancel"))
 
 
-def render_form(form, title, message='', formid=None, submit=__(u"Submit"), cancel_url=None, ajax=False, with_chrome=True, autosave=False):
+def render_form(form, title, message='', formid=None, submit=__(u"Submit"), cancel_url=None, ajax=False, with_chrome=True, action='', autosave=False, last_revision_id=None):
```
jace: Action parameter default should be `action=None`. The name `last_revision_id` is verbose. This could be `draft_revision` instead.

PR #211, opened 2019-02-13
Change to capture form submit.

vidya-ram:
Forms  can also use button with type `submit`.

### DIFF
--- a/baseframe/templates/baseframe/bootstrap3/forms.html.jinja2
+++ b/baseframe/templates/baseframe/bootstrap3/forms.html.jinja2
@@ -156,9 +156,10 @@
     <script type="text/javascript">
       $(function() {
         // Disable submit button when clicked. Prevent double click.
-        $('#{{ ref_id }} input[type="submit"]').click(function() {
-          $(this).addClass('submit-disabled').prop('disabled', true);
-          $('#{{ ref_id }}').find(".loading").removeClass('hidden');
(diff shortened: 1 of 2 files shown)


baseframe/templates/baseframe/bootstrap3/forms.html.jinja2
```diff
@@ -156,9 +156,10 @@
     <script type="text/javascript">
       $(function() {
         // Disable submit button when clicked. Prevent double click.
-        $('#{{ ref_id }} input[type="submit"]').click(function() {
-          $(this).addClass('submit-disabled').prop('disabled', true);
-          $('#{{ ref_id }}').find(".loading").removeClass('hidden');
+        $('#{{ ref_id }}').submit(function() {
+          $(this).find('button[type="submit"]').addClass('submit-disabled').prop('disabled', true);
+          $(this).find('input[type="submit"]').prop('disabled', true).addClass('submit-disabled');
```
jace: Any reason the chaining order is reversed between these two lines?

vidya-ram: Earlier it was handling form submits for input types as submit. Now added the case to handle forms submit with buttons of type submit.

jace:
This has broken Lastuser's authorize form. Since the Authorise and Deny buttons are disabled on click, the browser will not submit them, so Lastuser does not know whether the user authorised or declined.

PR #213, opened 2019-03-01
added AllowedIf and aliased RequiredIf

iambibhas:


baseframe/forms/validators.py
```diff
@@ -16,16 +16,33 @@
...
+    'StopValidation', 'URL', 'ValidationError']
+__all__ = __local + __imported
 
 
 EMAIL_RE = re.compile(r'\b[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,63}\b', re.I)
 
 
+class AllowedIf(object):
+    """
+    Validator that makes this field required if some other field is true.
```
jace: Validator that allows a value only if some other field has a value.

baseframe/forms/validators.py
```diff
@@ -16,16 +16,33 @@
...
+    """
+    def __init__(self, fieldname, message=None):
+        self.fieldname = fieldname
+        self.message = message or __("This field is not allowed if '{}' is not set".format(self.fieldname))
```
jace: We can't display "cfp_start_at" to the user. Extract the label and add it here. Also, you can't have a parameter inside `__()`. The format operation has to be outside.

baseframe/forms/validators.py
```diff
@@ -16,16 +16,33 @@
...
+
+    def __call__(self, form, field):
+        if field.data and not form[self.fieldname].data:
+            raise StopValidation(self.message)
+        elif not field.data and form[self.fieldname].data:
+            raise StopValidation(__("This is required"))
```
jace: The `elif` does not apply. This is `AllowedIf`, not `RequiredIf`.

baseframe/forms/validators.py
```diff
@@ -42,7 +59,7 @@ def __call__(self, form, field):
                 raise StopValidation(self.message)
 
 
-class OptionalIfNot(OptionalIf):
+class RequiredIf(OptionalIf):
     """
     Validator that makes this field optional if the value of some other field is false.
     """
```
jace: From the description and code, this it not a `RequiredIf`. Rename back to `OptionalIfNot`. We need a separate `RequiredIf`.

baseframe/forms/validators.py
```diff
@@ -16,23 +16,38 @@
...
+
+
 class OptionalIf(object):
     """
     Validator that makes this field optional if the value of some other field is true.
     """
     def __init__(self, fieldname, message=None):
         self.fieldname = fieldname
-        self.message = message or __("This is required")
+        self.message = message or __("This field is required")
```
jace: "field" is not something users understand. The earlier message is more appropriate.

baseframe/forms/validators.py
```diff
@@ -16,23 +16,38 @@
...
 EMAIL_RE = re.compile(r'\b[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,63}\b', re.I)
 
 
+class AllowedIf(object):
+    """
+    Validator that allows a value only if some other field has a value.
+    """
+    def __init__(self, fieldname, message=None):
+        self.fieldname = fieldname
+        self.message = message or __("This field is not allowed")
```
jace: The validator needs to indicate why it failed. This is not sufficient.

tests/fixtures.py
```diff
@@ -69,6 +69,16 @@ class OptionalIfNotFormTest(forms.Form):
     content = forms.TextAreaField(__("Content"))
 
 
+class RequiredIfFormTest(forms.Form):
+    blurb = forms.TextAreaField(__("Blurb"), validators=[forms.validators.RequiredIf('content')])
```
jace: In general, `RequiredIf` should reference earlier fields as that is how the user will see them.

tests/test_validators.py
```diff
@@ -129,6 +133,26 @@ def test_html_snippet_invalid_urls(self):
             self.all_urls_form.process(content_with_urls=snippet)
             self.assertEqual(self.all_urls_form.validate(), False)
 
+    def test_required_if(self):
+        self.required_if_form.process(content=u"Content", blurb=u"Blurb")
+        self.assertTrue(self.required_if_form.validate())
```
jace: We're using Pytest's `assert` statement in all new tests. Older tests can be migrated gradually.

baseframe/forms/validators.py
```diff
@@ -16,42 +17,79 @@
...
     """
-    Validator that makes this field optional if the value of some other field is true.
+    Validator that allows a value only if another field also has a value.
+
+    :param str fieldname: Name of the other field
+    :param str message: Validation error message. Will be formatted with an optional ``{field}}`` label
+    """
+    def __init__(self, fieldname, message=None):
+        self.fieldname = fieldname
+        self.message = message or __(u"This requires {field} to be specified")
```
iambibhas: Put quotes around the `field`? Otherwise this is how the error looks like -

> This requires Submissions open at to be specified

jace:
[WTForms has distinct `DataRequired` and `InputRequired` validators](https://wtforms.readthedocs.io/en/stable/validators.html#wtforms.validators.DataRequired) because of the problem with doing a `bool(value)` test on a value like `0`, which is technically input, but evaluates to false.

In this PR we are using `(None, '')` to specifically exclude `0`, but this is incomplete. It will fail with a [`SelectMultipleField`](https://wtforms.readthedocs.io/en/stable/fields.html#wtforms.fields.SelectMultipleField) where a blank value is `[]`. There is also the ambiguity of a checkbox field, where the possible values are `True` or `False`. Should these be included in the test as missing values?

baseframe/forms/validators.py
```diff
@@ -16,42 +17,79 @@
...
+        if field.data:
+            if form[self.fieldname].data in (None, ''):
+                raise StopValidation(self.message.format(field=form[self.fieldname].label.text))
+
+
+class OptionalIf(Optional):
+    """
+    Validator that makes this field optional if another field has data. If this
+    field is required when the other field is empty, chain it with
+    :class:`DataRequired`::
```
iambibhas: Doesn't `OptionalIf` name imply that it'll be Required otherwise? Same question with `RequiredIf`

jace:
If we take a whitelist approach, the zero values are `(0, 0.0, Decimal('0'))`. The empty data test can then be `value not in (0, 0.0, Decimal('0')) and not bool(value)`.

iambibhas:
Can we call `InputRequired` validator instead of doing the checks ourselves?

jace:
Pending question: should `False` be added to the zero values list in `is_empty`? We will have False in a form field when it's a checkbox that hasn't been checked. There are two ways to look at this:

1. As a checkbox, the "required" flag is meaningless. We can't force the user to check it, because then there's no reason to have it.

2. On the other hand, with `RequiredIf` there's a clear connection between checking a box and requiring an associated field to have a value. There is also the matter of terms of service acceptance forms where checking the box is indeed required as an act of user consent. Therefore `False` should not be treated as a zero value.

jace:
I'm going with _not_ including `False` in the zero value list for is_empty.

PR #214, opened 2019-03-05
Date and time field rendering

### LINKED ISSUE #144: Date and time field rendering
[Browser support for `date` and `time` input fields is still variable](http://caniuse.com/#feat=input-datetime). Baseframe offers `DateField` and `DateTimeField` that are affected by this. The field expects `YYYY-MM-DD` from the date field in a browser, but compliant browsers will render in the user's locale and translate when submitting (to be re-confirmed). Non-compliant browsers will render a simple text field. Unfortunately, this makes help descriptions tricky as we cannot know in advance how the field will render.

Baseframe should offer browser polyfills to ensure consistent rendering and help messages around these fields.

Related: hasgeek/funnel#182

vidya-ram:
1. Switch to using type="time"
2. Remove all JavaScript libraries that help with rendering time fields
3. Use placeholder="--:--" as a hint for Safari. Other browsers appear to ignore the placeholder.

Fixes #144


baseframe/forms/widgets.py
```diff
@@ -120,11 +120,11 @@ def __call__(self, field, **kwargs):
             value = ' '
         class_ = kwargs.pop('class', kwargs.pop('class_', ''))
         date_value, time_value = value.split(' ', 1)
-        return Markup(u'<input type="date" class="datetime-date %s" %s /> <input type="text" class="datetime-time %s" %s /> %s' % (
+        return Markup(u'<input type="date" class="datetime-date %s" %s /> <input type="time" class="datetime-time %s" %s /> %s' % (
             class_,
             wtforms.widgets.html_params(name=field.name, id=field_id + '-date', value=date_value, **kwargs),
             class_,
-            wtforms.widgets.html_params(name=field.name, id=field_id + '-time', value=time_value, **kwargs),
+            wtforms.widgets.html_params(name=field.name, id=field_id + '-time', value=time_value, placeholder=_("--:--"), **kwargs),
```
jace: `_()` is not required here as far as I can tell.

jace:
We can also drop the JS library from Baseframe's collection (in assets.py, and the file itself).

baseframe/forms/widgets.py
vidya-ram: Fixed

PR #215, opened 2019-03-22
added shortdate and longdate filters

iambibhas:


baseframe/filters.py
```diff
@@ -166,3 +167,25 @@ def cdata(text):
...
+        utc_now = datetime.now(utc)
+    else:
+        dt = value
+        utc_now = datetime.utcnow().date()
+    if dt > (utc_now - timedelta(days=int(current_app.config.get('SHORTDATE_THRESHOLD_DAYS', 0)))):
+        return dt.strftime('%e %b')
+    else:
+        # The string replace hack is to deal with inconsistencies in the underlying
+        # implementation of strftime. See https://bugs.python.org/issue8304
+        return six.text_type(dt.strftime('%e %B %Y')).replace(u"'", u"’")
```
jace: There is no `'` in the format string. Why is the string replace required?

baseframe/filters.py
```diff
@@ -166,3 +168,25 @@ def cdata(text):
...
+        utc_now = datetime.now(utc)
+    else:
+        dt = value
+        utc_now = request_timestamp().date()
+    if dt > (utc_now - timedelta(days=int(current_app.config.get('SHORTDATE_THRESHOLD_DAYS', 0)))):
+        return dt.strftime('%e %b')
+    else:
+        # The string replace hack is to deal with inconsistencies in the underlying
+        # implementation of strftime. See https://bugs.python.org/issue8304
+        return six.text_type(dt.strftime("%e %b '%y")).replace(u"'", u"’")
```
jace: The use of an apostrophe is likely to hurt localization in languages that don't use an apostrophe to represent abbreviations.

PR #216, opened 2019-03-25
Added cached localized country list

iambibhas:


baseframe/__init__.py
```diff
@@ -321,6 +323,19 @@ def get_timezone():
...
 
 
+def get_countries_list():
+    # Returns a localized list of country names and the ISO3166-2 code
+    return get_localized_countries(get_locale())
+
+
+@cache.memoize(timeout=86400)
+def get_localized_countries(locale):
+    pycountry_locale = gettext.translation('iso3166-2', pycountry.LOCALES_DIR, languages=[locale])
```
jace: Shouldn't this be `3166-1`? `-2` is for states.

baseframe/__init__.py
```diff
@@ -321,6 +323,19 @@ def get_timezone():
     return current_app.config.get('tz') or UTC
 
 
+def get_countries_list():
```
jace: `localized_country_list`, and a docstring is needed.

baseframe/__init__.py
```diff
@@ -321,6 +323,19 @@ def get_timezone():
...
+    # Returns a localized list of country names and the ISO3166-2 code
+    return get_localized_countries(get_locale())
+
+
+@cache.memoize(timeout=86400)
+def get_localized_countries(locale):
```
jace: `_localized_country_list_inner`

baseframe/__init__.py
```diff
@@ -321,6 +323,19 @@ def get_timezone():
...
+    pycountry_locale = gettext.translation('iso3166-2', pycountry.LOCALES_DIR, languages=[locale])
+    countries = [(country.alpha_2, pycountry_locale.gettext(country.name)) for country in pycountry.countries]
+    countries.sort()
```
jace: Same mistake as in Funnel: it's sorting by country code instead of name. Sort by localized name, then invert the list.

baseframe/__init__.py
iambibhas: I'm having trouble with this. pycountry only seems to be generating locales for `-2`. I'm tryinh to fix that.

tests/test_utils.py
```diff
@@ -0,0 +1,16 @@
...
+class TestUtils(TestCaseBaseframe):
+    def test_localized_country_list(self):
+        countries = _localized_country_list_inner('en')
+        assert dict(countries)['DE'] == u"Germany"
+        countries = _localized_country_list_inner('de')
+        assert dict(countries)['DE'] == u"Deutschland"
+        countries = _localized_country_list_inner('es')
+        assert dict(countries)['DE'] == u"Alemania"
+        countries = _localized_country_list_inner('hi')
+        assert dict(countries)['DE'] == u"जर्मनी"
```
jace: 1. You need a test for the outer function, which means you have to simulate locale by making a test request with a custom Accept header. Don't test the inner -- it's not meant to be used anywhere outside the outer function.

2. Test sorting by country name. Pick one country, say `DE` as above, and see whether it indexes before or after another country in various locales.

baseframe/__init__.py
```diff
@@ -321,6 +323,27 @@ def get_timezone():
     return current_app.config.get('tz') or UTC
 
 
+def localized_country_list():
```
jace: Docstring required.

baseframe/__init__.py
```diff
@@ -321,6 +323,27 @@ def get_timezone():
...
+    if locale == 'en':
+        countries = [(country.name, country.alpha_2) for country in pycountry.countries]
+    else:
+        pycountry_locale = gettext.translation('iso3166-1', pycountry.LOCALES_DIR, languages=[locale])
+        countries = []
+        for country in pycountry.countries:
+            name = pycountry_locale.gettext(country.name)
+            if six.PY2:
+                name = name.decode('utf-8')
+            countries.append((name, country.alpha_2))
```
jace: This version will be significantly faster because it (a) uses a list comprehension and (b) moves the `if` condition out of the loop:

```python
if six.PY2:
    countries = [(pycountry_locale.gettext(country.name).decode('utf-8'), country.alpha_2) for country in pycountry.countries]
else:
    countries = [(pycountry_locale.gettext(country.name), country.alpha_2) for country in pycountry.countries]
```

PR #217, opened 2019-04-19
Add emoji picker

vidya-ram:


baseframe/static/js/baseframe-material.js
```diff
@@ -58,7 +58,7 @@ function activate_widgets() {
   });
 
   // Activate codemirror on all textareas with class='stylesheet'
-  $('textarea.stylesheet').each(function() {
+  $('textarea.stylesheet:not([style*="display: none"])').each(function() {
```
jace: What is the reason for this exception to Codemirror? What broke? The `:not` selector is expensive, so maybe we should explicitly add a `codemirror` class or a `data-widget="codemirror"` attribute and look for it.

vidya-ram: For dynamic sub label forms in funnel, code mirror  has to be initialised again but that was causing already changed textarea to be initialised again and two code mirror instances were created. `:not` selector was to avoid initialising already initialised textarea.

vidya-ram: We have removed description field from label model, so I will revert this change.

PR #223, opened 2019-05-20
staticdata is no longer needed

### LINKED ISSUE #203: Replace the country list in staticdata with an external library
The `pycountry` library contains all the static data we want in Baseframe and apps: countries, languages and currencies. All uses of Baseframe's static list in apps should be replaced with `pycountry` (or a similar library) and the list should be removed here.

iambibhas:
Fixes #203.


jace:
There are two references that need to be fixed before this module can be removed. https://github.com/search?q=org%3Ahasgeek+staticdata&type=Code

iambibhas:
Github showed me just one instance, in hacknight. replaced that with localized country list.

PR #224, opened 2019-05-22
Use invisible recaptcha

### LINKED ISSUE #143: Invisible reCAPTCHA field
[Google has introduced an "Invisible reCAPTCHA" product that requires no user interaction](https://wptavern.com/google-launches-invisible-recaptcha). As the API is likely different, Baseframe should provide a field and widget that renders it in a form instead of waiting for WTForms to update.

iambibhas:
Fixes #143.

This essentially forces invisible recaptcha wherever RecaptchaField is used. On any client app that uses this, it's required to set this in the config - 
```
RECAPTCHA_DATA_ATTRS = {'callback': "onInvisibleRecaptchaSubmit", 'size': "invisible"}
```

### DIFF
--- a/baseframe/templates/baseframe/bootstrap3/autoform.html.jinja2
+++ b/baseframe/templates/baseframe/bootstrap3/autoform.html.jinja2
@@ -12,7 +12,7 @@
   {{ widget_ext_scripts(form) }}
   <script type="text/javascript">
     $(function() {
-      {{ widgetscripts(form, script=false) }}
+      {{ widgetscripts(form, script=false, form_ref_id=ref_id) }}
     });
   </script>
   <script src="{{ 'parsley.js'|ext_asset_url }}" type="text/javascript"></script>
--- a/baseframe/templates/baseframe/bootstrap3/autoform.html.jinja2
+++ b/baseframe/templates/baseframe/bootstrap3/autoform.html.jinja2
@@ -12,7 +12,7 @@
   {{ widget_ext_scripts(form) }}
   <script type="text/javascript">
     $(function() {
-      {{ widgetscripts(form, script=false, form_ref_id=ref_id) }}
+      {{ widgetscripts(form, script=false, ref_id=ref_id) }}
     });
   </script>
   <script src="{{ 'parsley.js'|ext_asset_url }}" type="text/javascript"></script>
--- a/baseframe/templates/baseframe/bootstrap3/forms.html.jinja2
+++ b/baseframe/templates/baseframe/bootstrap3/forms.html.jinja2
@@ -300,10 +300,9 @@
         key: {{ field.results_key|tojson }}
       });
     {%- elif field.type == 'RecaptchaField' and form_ref_id != '' %}
-      function onInvisibleRecaptchaSubmit(recaptcha_response) {
-        console.log(recaptcha_response);
-        // document.getElementById("{{ form_ref_id }}").submit();
-      }
+      window.onInvisibleRecaptchaSubmit = function (recaptcha_response) {
+        document.getElementById("{{ form_ref_id }}").submit();
+      };
       $('#{{ form_ref_id }}').submit(function (event) {
         event.preventDefault();
         if (typeof grecaptcha !== "undefiend" && grecaptcha.getResponse() === '') {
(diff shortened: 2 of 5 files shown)


iambibhas:
@vidya-ram there is a confusion between `formid` and `ref_id` in `renderform` macro. `formid` isn't used almost anywhere. Rather `ref_id` is dynamically generated and used as `id` for `<form>` elements.

jace:
`formid` is submitted back to the backend as the `form.id` variable. It is used when the same endpoint handles multiple forms, to identify which was submitted. It is only used in the frontend to create a hidden input field, with `name="form.id" value="{{ formid }}"`. As a value rather than an id, it is not useful in JavaScript.

You need `ref_id` here.

baseframe/templates/baseframe/mui/forms.html.jinja2
```diff
@@ -101,13 +101,26 @@
   {%- endif %}
 {%- endmacro %}
 
-{% macro rendersubmit(buttons=[], css_class="", tabindex='', cancel_url='', style='horiz', csrf_error=false) -%}
+{% macro rendersubmit(form, ref_id, buttons=[], css_class="", tabindex='', cancel_url='', style='horiz', csrf_error=false) -%}
```
jace: Changing the parameters will break a lot of code. Add this to the end as optional values. https://github.com/search?q=org%3Ahasgeek+rendersubmit&type=Code

baseframe/templates/baseframe/mui/forms.html.jinja2
```diff
@@ -101,13 +101,26 @@
...
+<script>
+function onInvisibleRecaptchaSubmit() {
+  event.preventDefault();
+  if (grecaptcha.getResponse() === '') {
+    grecaptcha.execute();
+  } else {
+    document.getElementById("{{ ref_id }}").submit();
+  }
+}
+</script>
```
vidya-ram: This function can be added to `widgetscripts` macro and form, ref_id need not be included as parameters in  rendersubmit

iambibhas:
Added the scripts to `widgetscript` and no longer modifying submit button.

baseframe/templates/baseframe/bootstrap3/autoform.html.jinja2
```diff
@@ -12,7 +12,7 @@
   {{ widget_ext_scripts(form) }}
   <script type="text/javascript">
     $(function() {
-      {{ widgetscripts(form, script=false) }}
+      {{ widgetscripts(form, script=false, form_ref_id=ref_id) }}
```
jace: Use `ref_id=ref_id`. Don't change the parameter name unless there is good reason to. If `ref_id` is ambiguous, it should be renamed to `form_ref_id` everywhere.

baseframe/templates/baseframe/bootstrap3/forms.html.jinja2
```diff
@@ -299,6 +299,19 @@
         autocomplete_endpoint: {{ field.autocomplete_endpoint|tojson }},
         key: {{ field.results_key|tojson }}
       });
+    {%- elif field.type == 'RecaptchaField' and form_ref_id != '' %}
+      function onInvisibleRecaptchaSubmit(recaptcha_response) {
+        console.log(recaptcha_response);
+        // document.getElementById("{{ form_ref_id }}").submit();
```
jace: Is this incomplete?

baseframe/templates/baseframe/mui/forms.html.jinja2
```diff
@@ -342,6 +342,19 @@
         autocomplete_endpoint: {{ field.autocomplete_endpoint|tojson }},
         key: {{ field.results_key|tojson }}
       });
+    {%- elif field.type == 'RecaptchaField' and form_ref_id != '' %}
+      window.onInvisibleRecaptchaSubmit = function (recaptcha_response) {
+        console.log(recaptcha_response);
```
jace: `console.log` is like `print`. Remove it before checking-in.

jace:
> On any client app that uses this, it's required to set this in the config -
>
> ```RECAPTCHA_DATA_ATTRS = {'callback': "onInvisibleRecaptchaSubmit", 'size': "invisible"}```

If this is required everywhere, why is this config? It should be just code.

PR #233, opened 2019-06-21
Filter and validator for handling lists

jace:


baseframe/forms/filters.py
```diff
@@ -72,6 +74,20 @@ def rstrip_inner(value):
...
+def strip_each(chars=unicode_extended_whitespace):
+    """
+    Strip whitespace and remove blank elements from each element in an iterable.
+    Falsy values are returned unprocessed
+
+    :param chars: If specified, strip these characters instead of whitespace
+    """
+    def strip_each_inner(value):
+        if value:
+            return [sline for sline in [line.strip(chars) for line in value] if sline]
```
iambibhas: Would using `return filter(None, [line.strip(chars) for line in value])` be a bit faster?

jace: Function calls are always slower.

PR #234, opened 2019-06-25
Detecting usermodel and lastuser for userselectfield

### LINKED ISSUE #199: Auto detect lastuser and usermodel when using UserSelectMultiField
In all the uses of `UserSelectMultiField()`, `lastuser` and `usermodel` are specified separately. They should be auto detected using `current_app`.

Ref: https://github.com/search?q=org%3Ahasgeek+UserSelectMultiField&type=Code

iambibhas:
Fixes #199.

### DIFF
--- a/baseframe/forms/fields.py
+++ b/baseframe/forms/fields.py
@@ -391,6 +391,7 @@ class UserSelectFieldBase(object):
             return [(u.userid, u.pickername, True) for u in self.data]
 
     def process_formdata(self, valuelist):
+        retval = super(UserSelectFieldBase, self).process_formdata(valuelist)
         userids = valuelist
         # Convert strings in userids into User objects
         users = []
@@ -411,6 +412,7 @@ class UserSelectFieldBase(object):
             else:
                 users = self.usermodel.all(userids=userids)
         self.data = users
+        return retval
 
 
 class UserSelectField(UserSelectFieldBase, StringField):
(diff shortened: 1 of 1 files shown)


baseframe/forms/fields.py
```diff
@@ -391,7 +391,6 @@ def iter_choices(self):
             return [(u.userid, u.pickername, True) for u in self.data]
 
     def process_formdata(self, valuelist):
-        retval = super(UserSelectFieldBase, self).process_formdata(valuelist)
```
iambibhas: This base class is a subclass of `object`. This is unnecessary. `process_formdata()` doesn't seem to need to return anything.

jace: If what you say is true, this line should have failed because `object` does not provide a `process_formdata`. The line works because this is a mixin class and some other class in the hierarchy actually does provide this method.

Please restore this line.

baseframe/forms/fields.py
```diff
@@ -412,7 +411,6 @@ def process_formdata(self, valuelist):
             else:
                 users = self.usermodel.all(userids=userids)
         self.data = users
-        return retval
```
jace: Generally not safe to make API assumptions. `__init__` explicitly does not return anything. Does WTForms explicitly say this method does not return anything? If not, don't discard the return value.

PR #239, opened 2019-08-27
Fixed widget for single user select field

iambibhas:


iambibhas:
Don't merge yet. Need to test a bit more.

Issue #165, opened 2017-09-26
Split Baseframe's third-party assets into a separate repo


jace:
Baseframe's third-party asset collection messes with the commit history and content of this repository: committers get credit for code that is not theirs, and the repo appears to be mostly JS and CSS when the original content is mostly Python.

Third-party assets should be split out into an independent repository named `baseframe_assets`, with baseframe retaining an import hook to make them available as `baseframe.assets`.

karthikb351:
Will you then use baseframe_assets as a git submodule inside baseframe?
That way you get a link between versions of baseframe and baseframe_assets and can safely update/rollback.

jace:
I'm leaning towards totally independent packages:

1. Git submodules are confusing because they link to specific releases. Every update to the submodule will require an update to the parent repo.
2. We don't want to package assets into a PyPI release. Anyone using the source release should be able to install the other package to get assets.
3. Private assets such as commercially-licensed icons can now be in a private repo, imported into `baseframe` via an app-specified import hook.

PR #246, opened 2019-11-13
added blueprint to baseframe-assets

iambibhas:


baseframe/__init__.py
```diff
@@ -248,6 +248,12 @@ def init_app(
         app.assets.register('css_all', css_all)
         app.register_blueprint(self, static_subdomain=subdomain)
 
+        try:
+            module = __import__('baseframe-assets')
```
jace: Asset name should be a parameter to the function here. There's no need for `__import__` otherwise.

jace:
Original ticket: #165.

baseframe/__init__.py
```diff
@@ -248,6 +248,14 @@ def init_app(
         app.assets.register('css_all', css_all)
         app.register_blueprint(self, static_subdomain=subdomain)
 
+        for module_name in app.config.get('PRIVATE_ASSETS', []):
```
jace: Don't use config as it's a statutory requirement for the client app and not a configuration issue. Take it as a parameter named `asset_modules` instead.

baseframe/__init__.py
```diff
@@ -248,6 +248,14 @@ def init_app(
...
+            try:
+                module = __import__(module_name)
+                module.blueprint.init_app_assets(app, assets)
+                app.register_blueprint(module.blueprint)
+            except ImportError:
+                continue
```
jace: Log an error here (app logger or Blueprint logger).

baseframe/__init__.py
```diff
@@ -137,6 +137,7 @@ def init_app(
         bundle_css=None,
         assetenv=None,
         theme='bootstrap3',
+        asset_modules=[]
```
jace: `=()` or `=None` (don't use a mutable as a default value)

baseframe/__init__.py
```diff
@@ -248,13 +249,13 @@ def init_app(
         app.assets.register('css_all', css_all)
         app.register_blueprint(self, static_subdomain=subdomain)
 
-        for module_name in app.config.get('PRIVATE_ASSETS', []):
+        for module_name in asset_modules:
             try:
                 module = __import__(module_name)
-                module.blueprint.init_app_assets(app, assets)
+                module.blueprint.init_app(app)
                 app.register_blueprint(module.blueprint)
```
jace: My sense that these should be registered as a subpath of `/_baseframe`, to avoid taking up additional top-level names (as the top-level is also used by profiles). However, it's unclear if Flask routing will handle this correctly, so test it before committing this change. The static subdomain is also necessary if the app is using it.

```python
app.register_blueprint(module.blueprint, url_prefix='/_baseframe', static_subdomain=subdomain)
```

iambibhas: Tried this. Doesn't seem to work as expected.

baseframe/__init__.py
```diff
@@ -248,6 +249,13 @@ def init_app(
         app.assets.register('css_all', css_all)
         app.register_blueprint(self, static_subdomain=subdomain)
 
+        for module_name in asset_modules:
+            try:
+                module = __import__(module_name)
+                module.blueprint.init_app(app, subdomain)
```
jace: Register the blueprint here. Let the module only register an asset.

PR #249, opened 2019-12-11
Serve ads.txt (resolves #248)

### LINKED ISSUE #248: Serve ads.txt
Like `robots.txt` and `humans.txt`, an `ads.txt` must be served when a site has advertising. Baseframe should serve this file, defaulting to a blank file.

jace:


jace:
After reviewing this a bit, this is the wrong solution. Unlike `humans.txt` and `robots.txt`, the contents of `ads.txt` are a property of the site installation, not the source repository. `ads.txt` may be served as a static file, but the static contents come from config, not the source tree. I'm therefore closing without merging.

Reference document for how `ads.txt` works: https://iabtechlab.com/ads-txt/

PR #253, opened 2020-01-13
Emoji validate

qurbat:


test_requirements.txt
```diff
@@ -3,3 +3,4 @@ coverage
 coveralls
 psycopg2
 urllib3
+emoji
```
iambibhas: This needs to go in the `requires` list inside `setup.py`. Not here.

baseframe/forms/validators.py
```diff
@@ -319,6 +320,19 @@ def __call__(self, form, field):
...
+class IsEmoji(object):
+    """
+    Validate whether the supplied string is an emoji.
+    """
+    def __init__(self, message=None):
+        self.message = message
+
+    def __call__(self, form, field):
+        if field.data in emoji.UNICODE_EMOJI:
+            return
```
iambibhas: No need for this `return`. Check for the invalid case and raise `ValidationError` if needed.

baseframe/forms/validators.py
```diff
@@ -9,6 +9,7 @@
 from fractions import Fraction
 import datetime
 import re
+import emoji
```
iambibhas: install `isort` and sort these imports. Or install `pre-commit` and let it do that.

baseframe/forms/validators.py
```diff
@@ -319,6 +320,15 @@ def __call__(self, form, field):
             return
         else:
             raise ValidationError(self.message)
+            
+class IsEmoji(object):
+
```
jace: Docstring required.

baseframe/forms/validators.py
```diff
@@ -319,6 +320,15 @@ def __call__(self, form, field):
...
+    def __init__(self, message=None):
+        self.message = message
```
jace: Fallback to a default message as in the other validators.

baseframe/forms/validators.py
```diff
@@ -319,6 +320,15 @@ def __call__(self, form, field):
             return
         else:
             raise ValidationError(self.message)
+            
+class IsEmoji(object):
```
jace: This validator was inserted between two validators that are related to each other. This one should have been before or after them. Move it.

jace:
This is also missing tests, so it reduces overall test coverage.

PR #255, opened 2020-01-22
Baseframe Emoji Validation

qurbat:
This PR adds the following:

 - ```emoji``` requirement to ```setup.py```
 - validation functionality to ```validators.py``` and ```fixtures.py```
 - test for validation to ```test_validators.py```

### DIFF
--- a/baseframe/forms/validators.py
+++ b/baseframe/forms/validators.py
@@ -302,17 +302,16 @@ class NotEqualTo(_Comparison):
 
 class IsEmoji(object):
     """
-    Validate that field.data is either empty or contains an emoji
-    through use of the ``emoji `` package.
+    Validate field to contain a single emoji
 
     :param message:
         Error message to raise in case of a validation error.
     """
     def __init__(self, message=None):
-        self.message = message or _(u"This is not a valid emoji.")
-        
+        self.message = message or _(u'This is not a valid emoji.')
+
     def __call__(self, form, field):
-        if field.data and field.data not in emoji.UNICODE_EMOJI:
+        if field.data not in emoji.UNICODE_EMOJI:
             raise ValidationError(self.message)
                 
 class IsPublicEmailDomain(object):
--- a/baseframe/forms/validators.py
+++ b/baseframe/forms/validators.py
@@ -302,7 +302,7 @@ class NotEqualTo(_Comparison):
 
 class IsEmoji(object):
     """
-    Validate field to contain a single emoji
+    Validate field to contain a single emoji.
 
     :param message:
         Error message to raise in case of a validation error.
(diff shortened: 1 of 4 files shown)


baseframe/forms/validators.py
```diff
@@ -299,7 +300,23 @@ class NotEqualTo(_Comparison):
     def compare(self, value, other):
         return value != other
 
+class IsEmoji(object):
+    """
+    Validate that field.data does not belong to a public email domain.
+    If the domain lookup fails and mxsniff raises ``MXLookupException``, this validator
+    will still pass, as we expect that most domains are not public email domains.
```
jace: Wrong docstring.

baseframe/forms/validators.py
```diff
@@ -299,7 +300,23 @@ class NotEqualTo(_Comparison):
...
+
+    :param message:
+        Error message to raise in case of a validation error.
+    """
+    def __init__(self, message=None):
+        self.message = message or _(u'This is not a valid emoji.')
```
jace: Message should be in double quotes. Our convention is to use double quotes for UI strings that need translation, and single quotes for tokens where the exact string must be preserved.

baseframe/forms/validators.py
```diff
@@ -299,7 +300,23 @@ class NotEqualTo(_Comparison):
...
 
+    def __call__(self, form, field):
+        if field.data != "":
```
jace: Just `if field.data:` should do.

baseframe/forms/validators.py
```diff
@@ -299,7 +300,23 @@ class NotEqualTo(_Comparison):
...
+            if not field.data in emoji.UNICODE_EMOJI:
```
jace: You can collapse these if statements into one: `if field.data and field.data not in emoji.UNICODE_EMOJI:`

tests/fixtures.py
```diff
@@ -57,6 +57,13 @@ class UrlFormTest(forms.Form):
         filters=[forms.filters.strip()],
     )
 
+class EmojiFormTest(forms.Form):
+    emoji = forms.StringField(
+        __("Emoji"),
+        validators=[
+        forms.validators.IsEmoji()
```
iambibhas: this doesn't seem black compliant. If you have black setup, format this file once.

tests/test_validators.py
```diff
@@ -42,6 +44,18 @@ def test_invalid_url(self):
             url = 'https://hasgeek'
             self.form.process(url=url)
             assert not self.form.validate()
+            
+    def test_valid_emoji(self):
+        with self.app.test_request_context('/'):
+            dat = u'\u2604'
```
iambibhas: You can also try giving an actual emoji as input to see if it works, like `😱`.

qurbat: Added! Seems to work.

baseframe/forms/validators.py
iambibhas: @qurbat this is not fixed.

qurbat: @iambibhas fixed now

PR #257, opened 2020-01-30
Introducing flask-babel2 changes

iambibhas:


setup.py
```diff
@@ -29,7 +29,7 @@
     'Flask-Assets',
     'Flask-WTF>=0.14',
     'Flask-Caching',
-    'Flask-BabelEx',
+    'flask-babel2',
```
jace: Should be `Flask-Babel2` to match the egg below.

Issue #260, opened 2020-02-10
Add self.default_message for IsEmoji and IsPublicEmailDomain validators


qurbat:
- Add self.default_message (instead of relying on ```message or _(u"Text")``` for these two validators
 - Use double underscores (```__```) instead of single (```_```)

PR #261, opened 2020-02-10
Update validators.py

qurbat:
#260

### DIFF
--- a/baseframe/forms/validators.py
+++ b/baseframe/forms/validators.py
@@ -126,9 +126,11 @@ class AllowedIf(object):
     :param str message: Validation error message. Will be formatted with an optional ``{field}}`` label
     """
 
+    default_message = __(u"This requires ‘{field}’ to be specified")
+
     def __init__(self, fieldname, message=None):
         self.fieldname = fieldname
-        self.message = message or __(u"This requires ‘{field}’ to be specified")
+        self.message = message or self.default_message
 
     def __call__(self, form, field):
         if field.data:
@@ -151,10 +153,12 @@ class OptionalIf(Optional):
     :param str message: Validation error message
     """
 
+    default_message = __("This is required")
+
     def __init__(self, fieldname, message=None):
         super(OptionalIf, self).__init__()
         self.fieldname = fieldname
-        self.message = message or __("This is required")
+        self.message = message or self.default_message
 
     def __call__(self, form, field):
         if not is_empty(form[self.fieldname].data):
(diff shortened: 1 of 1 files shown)


baseframe/forms/validators.py
```diff
@@ -126,9 +126,11 @@ class AllowedIf(object):
     :param str message: Validation error message. Will be formatted with an optional ``{field}}`` label
```
jace: Typo in `{field}}` (unrelated to your PR, just noticed it).

baseframe/forms/validators.py
```diff
@@ -309,9 +315,10 @@ class IsEmoji(object):
     :param message:
         Error message to raise in case of a validation error.
     """
+    default_message = __(u"This is not a valid emoji.")
```
jace: Remove trailing period in this message. The others don't have it. Also audit for other trailing periods. The rule is this:

1. If the message is a single statement, no trailing period.
2. If the message has two (or more) statements, use a period at the end of each statement except the last one.
3. If the message ends in other punctuation such as ! or ?, leave those in. _Only_ the last full stop is to be removed.

baseframe/forms/validators.py
```diff
@@ -379,7 +390,7 @@ def __call__(self, form, field):
         if diagnosis.code == 0:
             return
         else:
-            raise StopValidation(self.message or _(diagnosis.message))
+            raise StopValidation(self.message or __(diagnosis.message))
```
jace: This is at runtime, so it should use `_` (single underscore). The difference is this:

1. `_` (alias for `gettext`) does an immediate translation to whatever locale is detected from the request. However, if there is no request, it falls back to the default, which is `en`. This means it cannot be used at the module level or class level, as those statements are executed at program load time.

2. `__` (alias for `lazy_gettext`) creates a "lazy string". It is not an actual string, but a class that does the translation whenever it is cast into a string. Assuming that such casting is only done in templates, it is safe to create lazy strings at program load time. It is a safe substitute for `_`, but since it is relatively inefficient, it should not be used when `_` will suffice.

baseframe/forms/validators.py
```diff
@@ -402,11 +413,11 @@ class ValidUrl(object):
...
     def __init__(self, message=None, message_urltext=None, invalid_urls=[]):
-        self.message = message or _(
+        self.message = message or __(
             u'The URL “{url}” is not valid or is currently inaccessible'
         )
         self.invalid_urls = invalid_urls
-        self.message_urltext = message_urltext or _(
+        self.message_urltext = message_urltext or __(
             u'The URL “{url}” linked from “{text}” is not valid or is currently inaccessible'
         )
```
jace: In this case `__` is the correct choice as `__init__` is called at load time.

jace: Except these too should be on the class as `default_message` and `default_message_urltext`.

PR #270, opened 2020-03-20
datetime i18n 

qurbat:
- Registers filters for datetime i18n

### DIFF
--- a/baseframe/filters.py
+++ b/baseframe/filters.py
@@ -215,36 +215,39 @@ def longdate(value):
 
 
 @baseframe.app_template_filter('date')
-def date_filter(value, format='medium'):
+def date_filter(value, format='medium', usertz=True):
     if isinstance(value, datetime):
         if value.tzinfo is None:
             dt = UTC.localize(value).astimezone(get_timezone())
         else:
-            dt = value.astimezone(get_timezone())
-    else:
-        dt = value
-    return format_date(dt, format=format, locale=get_locale())  # NOQA
+            if usertz is True:
+                dt = value.astimezone(get_timezone())
+            else:
+                dt = value
+    return format_date(dt, format=format, locale=get_locale())  # NOQA: A002
 
 
 @baseframe.app_template_filter('format_time')
-def time_filter(value, format='short'):
+def time_filter(value, format='short', usertz=True):
     if isinstance(value, datetime):
         if value.tzinfo is None:
             dt = UTC.localize(value).astimezone(get_timezone())
         else:
-            dt = value.astimezone(get_timezone())
-    else:
-        dt = value
-    return format_time(dt, format=format, locale=get_locale())  # NOQA
+            if usertz is True:
+                dt = value.astimezone(get_timezone())
+            else:
+                dt = value
+    return format_time(dt, format=format, locale=get_locale())  # NOQA: A002
 
 
 @baseframe.app_template_filter('datetime')
-def datetime_filter(value, format='medium'):
+def datetime_filter(value, format='medium', usertz=True):
     if isinstance(value, datetime):
         if value.tzinfo is None:
             dt = UTC.localize(value).astimezone(get_timezone())
         else:
-            dt = value.astimezone(get_timezone())
-    else:
-        dt = value
-    return format_datetime(dt, format=format, locale=get_locale())  # NOQA
+            if usertz is True:
+                dt = value.astimezone(get_timezone())
+            else:
+                dt = value
+    return format_datetime(dt, format=format, locale=get_locale())  # NOQA: A002
--- a/baseframe/filters.py
+++ b/baseframe/filters.py
@@ -223,10 +223,7 @@ def date_filter(value, format='medium', locale=None, usertz=True):
             dt = value.astimezone(get_timezone())
     else:
         dt = value
-    if locale is not None:
-        return format_date(dt, format=format, locale=locale)  # NOQA: A002
-    else:
-        return format_date(dt, format=format, locale=get_locale())  # NOQA: A002
+    return format_date(dt, format=format, locale=locale if locale else get_locale())  # NOQA: A002
 
 
 @baseframe.app_template_filter('format_time')
@@ -238,10 +235,7 @@ def time_filter(value, format='short', locale=None, usertz=True):
             dt = value.astimezone(get_timezone())
     else:
         dt = value
-    if locale is not None:
-        return format_time(dt, format=format, locale=locale)  # NOQA: A002
-    else:
-        return format_time(dt, format=format, locale=get_locale())  # NOQA: A002
+    return format_time(dt, format=format, locale=locale if locale else get_locale())  # NOQA: A002
 
 
 @baseframe.app_template_filter('datetime')
@@ -253,7 +247,4 @@ def datetime_filter(value, format='medium', locale=None, usertz=True):
             dt = value.astimezone(get_timezone())
     else:
         dt = value
-    if locale is not None:
-        return format_datetime(dt, format=format, locale=locale)  # NOQA: A002
-    else:
-        return format_datetime(dt, format=format, locale=get_locale())  # NOQA: A002
+    return format_datetime(dt, format=format, locale=locale if locale else get_locale())  # NOQA: A002
--- a/baseframe/filters.py
+++ b/baseframe/filters.py
@@ -211,3 +211,42 @@ def longdate(value):
     else:
         dt = value
     return dt.strftime('%e %B %Y')
+
+
+@baseframe.app_template_filter('loclongdate')
+def loclongdate(value):
+    if isinstance(value, datetime):
+        if value.tzinfo is None:
+            dt = UTC.localize(value).astimezone(get_timezone())
+        else:
+            dt = value.astimezone(get_timezone())
+    else:
+        dt = value
+    return format_date(dt, format='long', locale=get_locale())
+
+
+@baseframe.app_template_filter('loctime')
+def loctime(value):
+    if isinstance(value, datetime):
+        if value.tzinfo is None:
+            dt = UTC.localize(value).astimezone(get_timezone())
+        else:
+            dt = value.astimezone(get_timezone())
+    else:
+        dt = value
+    return format_datetime(dt, "hh:mm a", locale=get_locale())
+
+
+@baseframe.app_template_filter('locdatetime')
+def locdatetime(value):
+    if isinstance(value, datetime):
+        if value.tzinfo is None:
+            dt = UTC.localize(value).astimezone(get_timezone())
+        else:
+            dt = value.astimezone(get_timezone())
+    else:
+        dt = value
+    locdate = format_date(dt, format='long', locale=get_locale())
+    loctime = format_time(dt, "h:mm a", locale=get_locale())
+    locdatetime = (str(locdate) + ' ' + str(loctime))
+    return locdatetime
--- a/baseframe/filters.py
+++ b/baseframe/filters.py
@@ -13,7 +13,8 @@ from coaster.gfm import markdown
 from coaster.utils import md5sum, text_blocks
 
 from . import b_ as _
-from . import baseframe, cache, current_app, get_timezone
+from . import baseframe, cache, current_app, get_locale, get_timezone
+from babel.dates import format_date, format_time, format_datetime
 from .utils import request_timestamp
 from .views import ext_assets
 
--- a/baseframe/filters.py
+++ b/baseframe/filters.py
@@ -218,9 +218,12 @@ def longdate(value):
 def date_filter(value, format='medium', usertz=True):
     if isinstance(value, datetime):
         if value.tzinfo is None:
-            dt = UTC.localize(value).astimezone(get_timezone())
+            if usertz:
+                dt = UTC.localize(value).astimezone(get_timezone())
+            else:
+                dt = value
         else:
-            if usertz is True:
+            if usertz:
                 dt = value.astimezone(get_timezone())
             else:
                 dt = value
@@ -233,9 +236,12 @@ def date_filter(value, format='medium', usertz=True):
 def time_filter(value, format='short', usertz=True):
     if isinstance(value, datetime):
         if value.tzinfo is None:
-            dt = UTC.localize(value).astimezone(get_timezone())
+            if usertz:
+                dt = UTC.localize(value).astimezone(get_timezone())
+            else:
+                dt = value
         else:
-            if usertz is True:
+            if usertz:
                 dt = value.astimezone(get_timezone())
             else:
                 dt = value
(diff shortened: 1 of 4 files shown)


baseframe/filters.py
```diff
@@ -211,3 +212,42 @@ def longdate(value):
...
+            dt = UTC.localize(value).astimezone(get_timezone())
+        else:
+            dt = value.astimezone(get_timezone())
+    else:
+        dt = value
+    return format_datetime(dt, "hh:mm a", locale=get_locale())
+
+
+@baseframe.app_template_filter('locdatetime')
+def locdatetime(value):
```
jace: Filter name `datetime` and function name `datetime_filter`

jace: This also accepts a `format='medium'` parameter.

baseframe/filters.py
```diff
@@ -211,3 +212,42 @@ def longdate(value):
...
+    if isinstance(value, datetime):
+        if value.tzinfo is None:
+            dt = UTC.localize(value).astimezone(get_timezone())
+        else:
+            dt = value.astimezone(get_timezone())
+    else:
+        dt = value
+    locdate = format_date(dt, format='long', locale=get_locale())
+    loctime = format_time(dt, "h:mm a", locale=get_locale())
+    locdatetime = str(locdate) + ' ' + str(loctime)
```
jace: Here you needed to use `format_datetime` instead of making separate calls and concatenating in a locale-unsafe manner.

baseframe/filters.py
```diff
@@ -211,3 +212,42 @@ def longdate(value):
...
+@baseframe.app_template_filter('loctime')
+def loctime(value):
+    if isinstance(value, datetime):
+        if value.tzinfo is None:
+            dt = UTC.localize(value).astimezone(get_timezone())
+        else:
+            dt = value.astimezone(get_timezone())
+    else:
+        dt = value
+    return format_datetime(dt, "hh:mm a", locale=get_locale())
```
jace: Use `format_time` and give it a `format` parameter.

baseframe/filters.py
```diff
@@ -178,36 +179,37 @@ def cdata(text):
     return Markup('<![CDATA[' + text.replace(']]>', ']]]]><![CDATA[>') + ']]>')
 
 
-@baseframe.app_template_filter('shortdate')
```
jace: You can't remove existing filters until these are _all_ gone. https://github.com/search?q=org%3Ahasgeek+shortdate&type=Code

baseframe/filters.py
```diff
@@ -178,36 +179,37 @@ def cdata(text):
...
-def shortdate(value):
+@baseframe.app_template_filter('date')
+def date_filter(value, fm='medium'):
```
jace: Call it `format` since you're simply passing this on. You will have to ignore a warning from PyFlakes.

baseframe/filters.py
```diff
@@ -178,36 +179,37 @@ def cdata(text):
...
+        if value.tzinfo is None:
+            dt = UTC.localize(value).astimezone(get_timezone())
+        else:
+            dt = value.astimezone(get_timezone())
     else:
-        # The string replace hack is to deal with inconsistencies in the underlying
-        # implementation of strftime. See https://bugs.python.org/issue8304
-        return six.text_type(dt.strftime("%e %b '%y")).replace(u"'", u"’")
+        dt = value
+    return format_time(dt, "hh:mm a", locale=get_locale())
```
jace: `format_time` also takes `format` as a preset string. Pass it in. Don't hardcode a rendering.

baseframe/filters.py
qurbat: I'll perform a code review and phase out usage of existing filters.

baseframe/filters.py
```diff
@@ -211,3 +212,39 @@ def longdate(value):
...
+@baseframe.app_template_filter('date')
+def date_filter(value, format='medium'):
+    if isinstance(value, datetime):
+        if value.tzinfo is None:
+            dt = UTC.localize(value).astimezone(get_timezone())
+        else:
+            dt = value.astimezone(get_timezone())
+    else:
+        dt = value
+    return format_date(dt, format=format, locale=get_locale())  # NOQA
```
jace: Use the specific `NOQA` code. For example, if it's an error `E001`, use `# noqa: E001`. This will prevent ignoring _other_ issues.

baseframe/filters.py
```diff
@@ -211,3 +212,39 @@ def longdate(value):
...
     return dt.strftime('%e %B %Y')
+
+
+@baseframe.app_template_filter('date')
+def date_filter(value, format='medium'):
+    if isinstance(value, datetime):
+        if value.tzinfo is None:
+            dt = UTC.localize(value).astimezone(get_timezone())
+        else:
+            dt = value.astimezone(get_timezone())
```
jace: This won't work for Vidya's use case where she wants to display a project's datetimes in the project's timezone, not the user's.

Try this: If the datetime has a timezone _other_ than UTC, don't change the timezone. However, this can be hard to check for as `pytz.UTC` is _not_ a singleton.

You can check for `tzoffset` being non-zero, but that too is iffy as British time is zero offset for part of the year. This will mean that a project in that timezone will appear in the user's timezone, while all other projects will appear in the local timezone.

The other option is to add an additional `usertz=True` parameter, so the caller can ask for having their tz preserved.

Finally, this will need unit tests to confirm the code behaves as expected in these various situations.

baseframe/filters.py
```diff
@@ -211,3 +212,39 @@ def longdate(value):
...
+    return format_date(dt, format=format, locale=get_locale())  # NOQA
+
+
+@baseframe.app_template_filter('format_time')
+def time_filter(value, format='short'):
+    if isinstance(value, datetime):
+        if value.tzinfo is None:
+            dt = UTC.localize(value).astimezone(get_timezone())
+        else:
+            dt = value.astimezone(get_timezone())
```
jace: Same timezone quandary here.

baseframe/filters.py
```diff
@@ -211,3 +212,39 @@ def longdate(value):
...
+    return format_time(dt, format=format, locale=get_locale())  # NOQA
+
+
+@baseframe.app_template_filter('datetime')
+def datetime_filter(value, format='medium'):
+    if isinstance(value, datetime):
+        if value.tzinfo is None:
+            dt = UTC.localize(value).astimezone(get_timezone())
+        else:
+            dt = value.astimezone(get_timezone())
```
jace: Here too.

baseframe/filters.py
```diff
@@ -211,3 +212,48 @@ def longdate(value):
     else:
         dt = value
     return dt.strftime('%e %B %Y')
+
+
+@baseframe.app_template_filter('date')
+def date_filter(value, format='medium', usertz=True):
+    if isinstance(value, datetime):
+        if value.tzinfo is None:
+            dt = UTC.localize(value).astimezone(get_timezone())
```
jace: Missed checking `usertz` here.

baseframe/filters.py
```diff
@@ -211,3 +212,48 @@ def longdate(value):
...
+        else:
+            if usertz is True:
```
jace: You don't need `is True`. Just `if usertz:` will do.

baseframe/filters.py
```diff
@@ -211,3 +212,57 @@ def longdate(value):
...
+                dt = UTC.localize(value).astimezone(get_timezone())
+            else:
+                dt = value
+        else:
+            if usertz:
+                dt = value.astimezone(get_timezone())
+            else:
+                dt = value
+    else:
+        dt = value
```
jace: This could be much simpler:
```python
if isinstance(value, datetime) and usertz:
    if value.tzinfo is None:
        dt = UTC.localize(value).astimezone(get_timezone())
    else:
        dt = value.astimezone(get_timezone())
else:
    dt = value
```

tests/test_filters.py
```diff
@@ -12,6 +12,7 @@ class TestDatetimeFilters(TestCaseBaseframe):
     def setUp(self):
         super(TestDatetimeFilters, self).setUp()
         self.now = utcnow()
+        self.today = date.today()
```
jace: Use a static date so that you can compare with a static string instead of the output of `strftime`.

tests/test_filters.py
```diff
@@ -86,6 +87,12 @@ def test_longdate_datetime_with_tz(self):
         with self.app.test_request_context('/'):
             assert filters.longdate(testdate) == testdate.strftime('%e %B %Y')
 
+    def test_date_filter(self):
+        testdate = self.today
+        testdate_string = self.today.strftime('%Y-%m-%d')
+        with self.app.test_request_context('/'):
+            assert filters.date_filter(testdate, 'yyyy-MM-dd', usertz=False) == testdate_string
```
jace: Test with (a) default format, (b) named formats (short, medium, long) and (c) custom format strings. Compare all with the static output string.

Then do this again with a different language passed to `test_request_context`.

baseframe/filters.py
jace: This is pending.

tests/test_filters.py
qurbat: Got it. Will push shortly.

tests/test_filters.py
```diff
@@ -12,6 +12,8 @@ class TestDatetimeFilters(TestCaseBaseframe):
     def setUp(self):
         super(TestDatetimeFilters, self).setUp()
         self.now = utcnow()
+        self.date = date(2020, 1, 1)
+        self.datetime = datetime(2020, 1, 1, 0, 0)
```
jace: Add `self.time` for tests.

jace: 1. Add a datetime with a UTC timezone. You currently have only a naive datetime
2. Add a datetime with a non-UTC timezone and render it with `usertz=False`

baseframe/filters.py
```diff
@@ -211,3 +212,48 @@ def longdate(value):
...
+        if value.tzinfo is None:
+            dt = UTC.localize(value).astimezone(get_timezone())
+        else:
+            dt = value.astimezone(get_timezone())
+    else:
+        dt = value
+    if locale is not None:
+        return format_date(dt, format=format, locale=locale)  # NOQA: A002
+    else:
+        return format_date(dt, format=format, locale=get_locale())  # NOQA: A002
```
jace: Reduce to one statement with `locale=locale if locale else get_locale()`

tests/test_filters.py
```diff
@@ -86,6 +88,38 @@ def test_longdate_datetime_with_tz(self):
...
+            assert filters.date_filter(self.date, "MMMM") == u'जनवरी'
+
+    def test_month_localized_en(self):
+        with self.app.test_request_context('/'):
+            assert filters.date_filter(self.date, "MMMM") == 'January'
+
+    def test_time_localized(self):
+        with self.app.test_request_context('/', headers={'Accept-Language': 'hi'}):
+            assert filters.time_filter(self.datetime, format='medium') == '12:00:00 am'
+
```
jace: Missing tests:

1. Passing in a `locale` parameter that differs from `Accept-Language` header.
2. Timezone rendering with or without `usertz` parameter.
3. Tests with a `time` object.
4. Datetimes with a timezone that is (a) in UTC, and (b) elsewhere.

tests/test_filters.py
```diff
@@ -86,6 +92,50 @@ def test_longdate_datetime_with_tz(self):
         with self.app.test_request_context('/'):
             assert filters.longdate(testdate) == testdate.strftime('%e %B %Y')
 
+    def test_date_filter(self):
+        with self.app.test_request_context('/'):
+            assert filters.date_filter(self.date, 'yyyy-MM-dd', usertz=False) == self.date.strftime('%Y-%m-%d')
+
+    def test_date_localized_short_hi(self):
+        with self.app.test_request_context('/', headers={'Accept-Language': 'hi'}):
+            assert filters.date_filter(self.date, format='short') == u'1/1/20'
```
jace: Problem with using Jan 1: can't tell if this is MDY or DMY. Under what conditions does Babel switch? Pick a test date > 12th of the month and add tests for MDY vs DMY vs YMD. Guessing this will change between `en_US` and `en_GB`/`en_IN`.

tests/test_filters.py
```diff
@@ -86,6 +92,50 @@ def test_longdate_datetime_with_tz(self):
...
+        with self.app.test_request_context('/', headers={'Accept-Language': 'hi'}):
+            assert filters.date_filter(self.date, "MMMM") == u'जनवरी'
+
+    def test_month_localized_en(self):
+        with self.app.test_request_context('/'):
+            assert filters.date_filter(self.date, "MMMM") == 'January'
+
+    def test_datetime_localized(self):
+        with self.app.test_request_context('/', headers={'Accept-Language': 'hi'}):
+            assert filters.time_filter(self.datetime, format='medium') == '12:00:00 am'
```
jace: Test time format `short` and `long` if those work.

baseframe/filters.py
```diff
@@ -211,3 +212,39 @@ def longdate(value):
...
+            dt = UTC.localize(value).astimezone(get_timezone())
+        else:
+            dt = value.astimezone(get_timezone())
+    else:
+        dt = value
+    return format_date(dt, format=format, locale=locale if locale else get_locale())  # NOQA: A002
+
+
+@baseframe.app_template_filter('format_time')
+def time_filter(value, format='short', locale=None, usertz=True):
```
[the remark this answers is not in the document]
jace: _If_ this is because `short` skips seconds while `medium` includes them, leave a comment explaining that. (We usually don't need to display seconds.)

tests/test_filters.py
```diff
@@ -86,6 +92,50 @@ def test_longdate_datetime_with_tz(self):
...
+
+    def test_time_localized(self):
+        with self.app.test_request_context('/', headers={'Accept-Language': 'hi'}):
+            assert filters.time_filter(self.time, format='full') == u'11:59:59 pm समन्वित वैश्विक समय'
```
jace: This one in English as well, just so the full text is on record.

baseframe/filters.py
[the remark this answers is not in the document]
qurbat: Yes, I thought ```short``` should be the default due to how time is usually stylized.

tests/test_filters.py
```diff
@@ -86,6 +92,70 @@ def test_longdate_datetime_with_tz(self):
...
+        with self.app.test_request_context('/'):
+            assert filters.datetime_filter(self.datetimeEST, format='full', usertz=True) == u'Friday, January 31, 2020 at 4:56:00 AM Coordinated Universal Time'
+
+    def test_date_dmy(self):
+        with self.app.test_request_context('/'):
+            assert filters.date_filter(self.datetime, format='short', locale='en_GB') == u'31/01/2020'
+
+    def test_date_mdy(self):
+        with self.app.test_request_context('/'):
+            assert filters.date_filter(self.datetime, format='short', locale='en_US') == u'1/31/20'
```
jace: The locale should be through the Accept Language header. We're clearly missing language variants here.

PR #289, opened 2020-05-12
Convert tags into buckets when statsd doesn't support it

jace:


baseframe/statsd.py
```diff
@@ -88,22 +86,29 @@ def init_app(self, app):
...
+            tags['app'] = current_app.config['SITE_ID']
             tag_sep = current_app.config['STATSD_TAGS']
+            tag_join = '='
+        else:
+            prefix = 'flask_app.{}'.format(current_app.config['SITE_ID'])
+            tag_sep = tag_join = '.'
+        if tags:
             name += tag_sep + tag_sep.join(
-                '='.join((str(t), str(v))) if v is not None else str(t)
+                tag_join.join((str(t), str(v))) if v is not None else str(t)
```
iambibhas, line 108: `tag_join` is undefined in the `else` block above.

iambibhas: oh never mind. Just saw it.

PR #291, opened 2020-05-13
Update validators.py

qurbat:
Add `https://`  validator

### DIFF
--- a/baseframe/forms/validators.py
+++ b/baseframe/forms/validators.py
@@ -41,7 +41,7 @@ __local = [
     'AllUrlsValid',
     'IsEmoji',
     'IsNotPublicEmailDomain',
-    'IsNotHttps',
+    'IsHttps',
     'IsPublicEmailDomain',
     'NoObfuscatedEmail',
     'AllowedIf',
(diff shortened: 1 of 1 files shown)


baseframe/forms/validators.py
```diff
@@ -397,6 +398,21 @@ def __call__(self, form, field):
...
+    """
+
+    default_message = __(u'The URL provided is not HTTPS')
+
+    def __init__(self, message=None, message_urltext=None):
+        self.message = message or self.default_message
+
+    def __call__(self, form, field):
+        if field.data:
+            if "https://" not in field.data:
```
jace, line 413: This will pass if the URL has `https://` in the middle. You want `field.data.startswith`.

baseframe/forms/validators.py
```diff
@@ -397,6 +398,21 @@ def __call__(self, form, field):
 # Legacy name
 ValidEmailDomain = ValidEmail
 
+class IsNotHttps(object):
```
jace, line 401: As per what it does, the validator should be named `IsHttps`.

baseframe/forms/validators.py
```diff
@@ -397,6 +398,21 @@ def __call__(self, form, field):
...
+    def __init__(self, message=None, message_urltext=None):
+        self.message = message or self.default_message
+
+    def __call__(self, form, field):
+        if field.data:
+            if "https://" not in field.data:
+                raise ValidationError(self.message)
+
 
 class ValidUrl(object):
```
jace, line 417: IMO, a robust implementation will add a `schemes` parameter to ValidUrl and allow any of those schemes to pass.

jace:
Replaced by #296.

PR #295, opened 2020-05-15
Use unicode literals and default messages

jace:


tests/test_filters.py
```diff
@@ -233,15 +235,15 @@ def test_avatar_url(self):
...
+            '//www.gravatar.com/avatar/00000000000000000000000000000000?d=mm',
         )
 
         # testing what if the user has an avatar already
         self.test_user.set_avatar(self.test_avatar_url)
         avatar_url = filters.avatar_url(self.test_user, self.test_avatar_size)
         self.assertEqual(
             avatar_url,
-            self.test_avatar_url + '?size=' + u'x'.join(self.test_avatar_size),
+            self.test_avatar_url + '?size=' + 'x'.join(self.test_avatar_size),
```
iambibhas, line 246: This function is named `test_avatar_url` as well. Maybe it should be something else?

PR #296, opened 2020-05-16
Expanded capabilities in ValidUrl

jace:
* Support a list of allowed schemes and domains
* Optionally disable URL tester
* Use html5lib instead of lxml for parsing safety
* Misc linter fixes

### DIFF
--- a/baseframe/forms/form.py
+++ b/baseframe/forms/form.py
@@ -176,7 +176,7 @@ class Form(BaseForm):
             form_validation_success.send(self)
 
     def errors_with_data(self):
-        # Convert lazy_gettext error strings into str so they don't cause problems downstream
+        # Convert lazy_gettext error strings into unicode so they don't cause problems downstream
         # (like when pickling)
         return {
             name: {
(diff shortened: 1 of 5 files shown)


baseframe/forms/form.py
```diff
@@ -176,7 +176,7 @@ def send_signals(self):
             form_validation_success.send(self)
 
     def errors_with_data(self):
-        # Convert lazy_gettext error strings into unicode so they don't cause problems downstream
+        # Convert lazy_gettext error strings into str so they don't cause problems downstream
```
jace, line 179: Doc fix doesn't belong in this PR. Slipped through.

tests/test_validators.py
```diff
@@ -197,6 +197,250 @@ def test_nonce_form_on_failure(self):
...
+            url = forms.StringField(
+                "URL",
+                validators=[
+                    forms.validators.ValidUrl(allowed_domains=lambda: ('example.net',))
+                ],
+            )
+
+        form = UrlForm(meta={'csrf': False})
+        form.url.data = 'http://www.example.com'
+        assert form.validate() is False
```
iambibhas, line 362: add a test for `True` as well maybe.

jace: That's the next test.

PR #299, opened 2020-05-25
Add initials filter

qurbat:
Add filter to return initials (e.g., `John Doe` > `JD`)

### DIFF
--- a/baseframe/filters.py
+++ b/baseframe/filters.py
@@ -55,6 +55,20 @@ def age(dt):
         return _("%(num)s years ago", num=int(delta.days / 365))
 
 
+@baseframe.app_template_filter('initials')
+def initials(field):
+    """
+    Return first and last initials from the given input, meant for use as avatar stand-in.
+    """
+    parts = field.split()
+    if len(parts) > 1:
+        return parts[0][0] + parts[-1][0]
+    elif parts[0]:
+        return parts[0][0]
+    else:
+        return ''
+
+
 @baseframe.app_template_filter('usessl')
 def usessl(url):
     """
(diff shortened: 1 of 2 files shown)


baseframe/filters.py
```diff
@@ -54,6 +56,19 @@ def age(dt):
     else:
         return _("%(num)s years ago", num=int(delta.days / 365))
 
+@baseframe.app_template_filter('initials')
+def initials(field):
+    """
+    Return first and last initials from the given input, meant for use as avatar stand-in.
+    """
+    parts = field.translate(str.maketrans('', '', string.punctuation)).split()
```
jace, line 64: This may be too elaborate for our needs.

tests/test_filters.py
```diff
@@ -203,6 +203,19 @@ def setUp(self):
...
+        self.assertEqual(initial, 'AE')
+
+        initial = filters.initials('A Slightly Longer Named Example')
+        self.assertEqual(initial, 'AE')
+
+        initial = filters.initials('Example')
+        self.assertEqual(initial, 'E')
+
+        initial = filters.initials('एक एक्साम्पल')
+        self.assertEqual(initial, 'एए')
```
jace, line 217: Needs tests for strings beginning and ending with a space.

baseframe/filters.py
```diff
@@ -54,6 +54,19 @@ def age(dt):
...
+    """
+    Return first and last initials from the given input, meant for use as avatar stand-in.
+    """
+    parts = field.split()
+    if len(parts) > 1:
+        return parts[0][0] + parts[-1][0]
+    elif parts[0]:
+        return parts[0][0]
+    else:
+        return ''
```
jace, line 68: Need a test that hits this line. Inputs: `None`, `''` (empty string) and `' '` (space). `None` will break this filter, so your very first line has to be `if not field: return ''`

PR #300, opened 2020-05-29
Add getInitials function

vidya-ram:


baseframe/static/js/baseframe-bootstrap.js
```diff
@@ -406,6 +406,24 @@ window.Baseframe.MapMarker.prototype.getDefaultLocation = function () {
   return { latitude: latitude, longitude: longitude };
 };
 
+window.Baseframe.Utils = {
+  getInitials: function (name) {
+    if (name) {
+      var parts = name.split(' ');
```
jace, line 412: We may have a name using `First\tLast`, so the split has to be for all whitespace characters. There is no validator preventing such whitespace characters in the username.

baseframe/static/js/baseframe-material.js
```diff
@@ -403,6 +403,24 @@ window.Baseframe.MapMarker.prototype.getDefaultLocation = function () {
   return { latitude: latitude, longitude: longitude };
 };
 
+window.Baseframe.Utils = {
+  getInitials: function (name) {
+    if (name) {
+      var parts = name.split(' ');
```
jace, line 409: Same here.

PR #303, opened 2020-06-10
Update line heights and font sizes

vidya-ram:


baseframe/static/sass/mui/_globals.scss
```diff
@@ -15,7 +15,7 @@
     font-family: $mui-base-font-family;
     font-size: $mui-base-font-size;
     font-weight: $mui-base-font-weight;
-    line-height: $mui-base-line-height;
+    line-height: $mui-base-line-height-px;
```
jace, line 18: How does this work? Does this remove `px` from the variable?

vidya-ram: Added a new line height variable in `px` for uniformity since the `$mui-base-line-height` is a ratio that is used in calculations in other places.
https://github.com/hasgeek/baseframe/pull/303/files#diff-f0cee5be043f68b99267dd6fad61a009R42

PR #304, opened 2020-06-11
Added timestamp filter

iambibhas:


tests/test_filters.py
```diff
@@ -186,6 +186,10 @@ def test_date_mdy(self):
                 == '1/31/20'
             )
 
+    def test_timestamp(self):
+        with self.app.test_request_context('/'):
+            assert filters.timestamp_filter(self.datetime) == 1580428800.0
```
jace, line 191: I don't think the microsecond precision is necessary or even suitable for sharing in public. It should be optional, with the filter taking `ms=False` or `microseconds=False` as default.

PR #305, opened 2020-06-17
render_form should accept layout template as a parameter

vidya-ram:


baseframe/forms/auto.py
```diff
@@ -47,6 +47,7 @@ def render_form(
     action=None,
     autosave=False,
     draft_revision=None,
+    template='',
```
jace, line 50: Should be `layout_template`

baseframe/static/js/baseframe-material.js
```diff
@@ -60,6 +60,16 @@ function activate_widgets() {
...
+  $('.js-show-password').click(function (argument) {
+    $(this).parent().find('.password-toggle').toggleClass('mui--hide');
+    $(this).parent().find('input').attr('type', 'text');
+  });
+
+  $('.js-hide-password').click(function (argument) {
+    $(this).parent().find('.password-toggle').toggleClass('mui--hide');
+    $(this).parent().find('input').attr('type', 'password');
+  });
 }
```
jace, line 73: This is adding to global JS. It should only be in the template's JS.

baseframe/forms/auto.py
```diff
@@ -56,9 +57,12 @@ def render_form(
         if isinstance(field.widget, wtforms.widgets.FileInput):
             multipart = True
     if not with_chrome:
-        template = THEME_FILES[current_app.config['theme']]['ajaxform.html.jinja2']
+        if not layout_template:
+            layout_template = THEME_FILES[current_app.config['theme']][
+                'ajaxform.html.jinja2'
+            ]
```
jace, line 63: Wait, `ajaxform.html.jinja2` is not a layout template. It's a regular template. `formlayout.html.jinja2` is the layout template.

PR #312, opened 2020-08-05
Support key rotation for secret keys

jace:
Depends on hasgeek/coaster#281

### DIFF
--- a/baseframe/__init__.py
+++ b/baseframe/__init__.py
@@ -198,8 +198,10 @@ class BaseframeBlueprint(Blueprint):
         # Setup secret key rotation
         if app.config['SECRET_KEY']:  # Always present as it's a default
             if not app.config.get('SECRET_KEYS'):
+                app.logger.debug("Setting SECRET_KEYS from SECRET_KEY")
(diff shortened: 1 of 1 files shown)


baseframe/__init__.py
```diff
@@ -194,6 +195,16 @@ def init_app(
                 ],
             )
 
+        # Setup secret key rotation
+        if app.config['SECRET_KEY']:  # Always present as it's a default
+            if not app.config.get('SECRET_KEYS'):
+                app.config['SECRET_KEYS'] = [app.config['SECRET_KEY']]
+        elif 'SECRET_KEYS' in app.config:
+            app.config['SECRET_KEY'] = app.config['SECRET_KEYS'][0]
```
iambibhas, line 203: This will fail and throw a confusing error, if `SECRET_KEYS = []` and `SECRET_KEY = ''` for some reason. It shouldn't happen, just mentioning.

jace: Changed to check for truthiness, and also log a DEBUG explaining which var is being set from which.

PR #314, opened 2020-08-06
Added cleanurl filter from hasjob

iambibhas:


jace:
Maybe move the new version from Funnel here?

jace (approved):

jace (approved):

PR #319, opened 2020-09-04
importing newrelic config file if it exists

iambibhas:


baseframe/__init__.py
```diff
@@ -404,6 +404,15 @@ def init_app(
...
         app.config['RECAPTCHA_DATA_ATTRS'].setdefault('size', 'invisible')
 
+        try:
+            import newrelic.agent
+            import os.path
+
+            if os.path.isfile('newrelic.ini'):
+                newrelic.agent.initialize('newrelic.ini')
+        except ImportError:
+            pass
```
jace, line 414: Needs `app.logger.debug` to alert dev that newrelic init has succeeded or failed.

PR #328, opened 2020-11-12
Move Recaptcha validator into Baseframe

jace:
Improved error handling for intermittent network problems.


jace:
Missing a test. The original test in Flask-WTF needs to be rewritten as it mocks urlopen but we use requests here.

PR #333, opened 2020-12-01
Moved csrf_token function to jinja env

iambibhas:


jace:
This is now part of #331 via 129264ee896cff063708abf89788049139acd1f7.

PR #343, opened 2021-02-22
Fixed emoji check in emoji validator

iambibhas:
From version 1.0.0, the `emoji` package changed the way you can check for existence of an emoji in their dicts. This is why our label tests were breaking. This fixes that. 

ref: https://github.com/carpedm20/emoji/issues/155


jace (approved):

jace:
Also change the minimum version in setup.py.

PR #360, opened 2021-05-20
Fix DateTimeField and some extras

jace:
* Fixed a bug in DateTimeField where it failed when time was not specified
* Replaced the datetime parser with dateutil for more flexibility in formats
* Added tests for various formats
* Replaced UI widget with `datetime-local`, with the option of a Firefox/Safari polyfill restoring separate `date` and `time` fields (not included in this PR, but backend will handle)
* Added `for` attribute to `label` tags in widget renderers
* Import Markup from markupsafe to stop DeprecationWarning from jinja2

Pending: CSS for `input[type=datetime-local]` to prevent full-width widget


jace:
A CSS fix is pending because `input[type=datetime-local]` currently renders full-width with wasted space:

image

This style comes from `.mui-textfield>input { width: 100% }`

jace:
Browser issue with `datetime-local` widget in Chrome:

image

This widget looks like a blank month, with the rest of the date present (caused by pressing backspace within the field).

**What should happen:** Alert the user the date is malformed, ask to fix before submit.

**What actually happens:** Field is submitted as blank, removing the entire value. This happens if any part is removed, including the time.

tests/test_fields.py
```diff
@@ -192,3 +200,93 @@ def test_escaped_label_text(self):
             label(**{'for': 'bar'}),
             """<label for="bar">&lt;script&gt;alert(&#34;test&#34;);&lt;/script&gt;</label>""",
         )
+
+
+class TestDateTimeField(BaseTestCase):
```
jace, line 205: Remove test since it's reproduced below

PR #364, opened 2021-06-09
Fixed redirect handler for ValidUrl

iambibhas:


baseframe/forms/validators.py
```diff
@@ -527,7 +527,7 @@ def check_url(
                 # 30 is the default redirect limit in `requests`
                 for _count in range(30):
                     r = requests.get(
-                        rurl,
+                        str(rurl),
```
iambibhas, line 530: Had to cast to `str` because mypy was complaining - 

> baseframe/forms/validators.py:530: error: Argument 1 to "get" has incompatible type "Optional[str]"; expected "Union[str, bytes]"

iambibhas: Need to handle that actually, instead of casting to `str`.

jace (approved):

PR #428, opened 2023-07-19
Check if Matomo URL is ending with '/'

djamg:


src/baseframe/blueprint.py
```diff
@@ -141,6 +141,9 @@ def init_app(
             raise ValueError("App has misconfigured secret keys")
         app.session_interface = RotatingKeySecureCookieSessionInterface()
 
+        # Check if Matomo URL is ending with '/' or else add it
+        if app.config.get('MATOMO_URL') and not app.config['MATOMO_URL'].endswith('/'):
+            app.config['MATOMO_URL'] += '/'
         # Default .js and tracking file for Matomo
         if app.config.get('MATOMO_URL') and app.config.get('MATOMO_ID'):
             app.config.setdefault('MATOMO_JS', 'matomo.js')
```
jace, line 149: The double `if` condition can be avoided:

```python
if app.config.get('MATOMO_URL') and app.config.get('MATOMO_ID'):
    app.config.setdefault('MATOMO_JS', 'matomo.js')
    app.config.setdefault('MATOMO_FILE', 'matomo.php')
    if not app.config['MATOMO_URL'].endswith('/'):
        app.config['MATOMO_URL'] += '/'
```

jace (changes requested):

jace (approved):

